PR #18, opened 2018-05-27
Implement Edmonds-Karp algorithm

jsbean:
The [Edmonds-Karp algorithm](https://en.wikipedia.org/wiki/Edmonds%E2%80%93Karp_algorithm) finds the maximum flow (and therefore minimum cut) of a [FlowNetwork](https://en.wikipedia.org/wiki/Flow_network).

We can use this minimum cut to partition a pitch spelling graph into two subgraphs. The source-side- and sink-side-ness can then be assigned to the nodes of the graph. Ultimately, we should be able to decode this info into sharps and flats and so on.


jsbean:
@brianlheim, this could be of interest!

PR #36, opened 2018-06-16
[PitchSpeller] Capacities

bwetherfield:
FIXME added for filling in edge capacities. Actual organization of this TBD.


jsbean:
Nice! I am going to point this to merge into the `wetherfield-speller` branch, instead of `master`.

Sources/PitchSpeller/Wetherfield/Wetherfield.swift
```diff
@@ -110,11 +110,17 @@ extension Graph where Value == Int {
         self.init([source, sink] + internalNodes)
         for node in internalNodes {
             insertEdge(from: source, to: node, value: 1)
+            // FIXME: insertEdge(from: source, to: node, value: epsilon *
```
jsbean: Thanks @bwetherfield! Apologies for my continued illiteracy: is this `lookupTable` supposed to be filled already (for the purpose of initializing the `Graph`), or is this supposed to be done in the process of manipulating an already-initialized `Graph`. What form is this `lookupTable` supposed to take here?

Also, is the `bigM` here arbitrarily large (in our case, `Double.greatestFiniteMagnitude`), or is it parameterized (i.e., customizable by the user)?

jsbean:
Do you want to take a stab at implementing this stuff, or do you want me to tackle it? It should be quick for me, but if you want some reps with the language, that also works!

Sources/PitchSpeller/Wetherfield/Wetherfield.swift
```diff
@@ -110,9 +110,17 @@ extension Graph where Value == Int {
...
+            // lookupTable[source.pitchclass, source.index, node.pitchclass, node.index])
             insertEdge(from: node, to: sink, value: 1)
+            // FIXME: insertEdge(from: node, to: sink, value: epsilon *
+            // lookupTable[node.pitchclass, node.index, sink.pitchclass, sink.index])
             for other in internalNodes.lazy.filter({ $0 != node }) {
                 insertEdge(from: node, to: other, value: 1)
+                // FIXME: if node.offset == other.offset && node.index == 1 {
```
jsbean: Given this requirement, it might make sense to shove in the pairs of nodes wrapped up together. Then we don't need to ask whether they represent the same item.

bwetherfield: agreed

Sources/PitchSpeller/Wetherfield/Wetherfield.swift
bwetherfield: I'm working on a bigM branch to handle this stuff!

lookupTable should be a big array of real values from which we instantiate the edge weights for our graph (later we need to add scale factors like proximity in score, time in score etc)

Sources/PitchSpeller/Wetherfield/EdgeCapacity.swift
```diff
@@ -0,0 +1,59 @@
+//
+//  EdgeCapacity.swift
+//  NotationModel
+//
+//  Created by Ben Wetherfield on 06/16/18.
+//
+
+enum EdgeCapacity:
```
jsbean: Ahh, nice! I avoided making `Edge.value` generic because "you ain't gonna need it. But here we are. Excited to see where this is going. If you wanna try, give a look at `Graph.Edge` in `Graph.swift`. The `value` instance property is currently a `Double`, but we could make `Edge` generic over some `Value` type. We would want to be thoughtful about the generic requirements of the `Value` type (`Hashable`, `Comparable`, `Numeric`, etc).

Sources/PitchSpeller/Wetherfield/EdgeCapacity.swift
```diff
@@ -13,6 +13,7 @@ enum EdgeCapacity:
 }
 
 // MARK: - Comparable
+// bigM edges always 'win' comparisons - otherwise like a regular comparison
```
jsbean: I am liking this, though I wonder what the difference between this and using `Double.infinity` would be.

bwetherfield: None! Is that a thing ? (whoops!)

jsbean: 👍

Sources/PitchSpeller/Wetherfield/EdgeCapacity.swift
jsbean: As of Swift 4.2, `Hashable` conformity comes for free (if the composed types are also `Hashable`, which implies (all?) primitive types). Right now, at least the `Node` values are stored in a `Dictionary` (roughly a `HashMap`) as keys. We don't necessarily need it for the `Edge` values, but in case we want `O(1)` access to 'em at some point, it could be worth investigating.

Luckily, it should only have to take adding `: Hashable`, and its there.

bwetherfield:
@jsbean, shall we repurpose this PR as the filling-in-edge-values task? I think that may have been the point initially but I got off track with all the attempts at implementing bigM (without using `.infinity` or `.max` properties)

jsbean:
@bwetherfield, sure! Go for it. Let me know if I can contribute to it, or if you want to take charge on it.

PR #42, opened 2018-06-17
[PitchSpeller / WIP / Experimental] Intedges

bwetherfield:
Alternative approach to edges using both integers and bigM. Potentially best of both worlds (cf. https://github.com/dn-m/NotationModel/pull/36). In particular makes != 0 comparisons foolproof for free.


Sources/PitchSpeller/Data Structures/Graph.swift
```diff
@@ -29,11 +29,15 @@ public struct Graph <Value: Hashable>: Hashable {
 
         public let source: Node
         public let destination: Node
-        public var value: Double
+        public var value: EdgeCapacity
+        
+        public var hashValue: Int {
```
jsbean: I'd say except for exceptional circumstances, we needn't implement our own `hashValue`. If all of the stored types here are themselves `Hashable`, we should get `Hashable` for free. Was the compiler whining that `Graph` no longer conformed to `Hashable`? Either we can relax the requirement for `Graph` to be `Equatable`, rather than `Hashable`, but considering we get it for free…

Sources/PitchSpeller/Wetherfield/EdgeCapacity.swift
```diff
@@ -0,0 +1,111 @@
...
+
+public enum EdgeCapacity:
+    Comparable,
+    Equatable,
+    Hashable
+{
+    case real(UInt)
+    case bigM
+    
+    public var hashValue: Int {
```
jsbean: Ditto re: `Hashable`. `struct` and `enum` values whose comprising values are `Hashable` get it for free.

Sources/PitchSpeller/Data Structures/Graph.swift
```diff
@@ -29,11 +29,15 @@ public struct Graph <Value: Hashable>: Hashable {
 
         public let source: Node
         public let destination: Node
-        public var value: Double
+        public var value: EdgeCapacity
```
jsbean: Instead of directly changing the concrete type of `Edge.value`, I'd suggest making it generic (over [`Numeric`](https://developer.apple.com/documentation/swift/numeric), etc.). Otherwise, we are tying this `Graph` implementation specifically to this use case.

jsbean:
Similar to the `FloatingPoint.infinity`, what about using [`UInt.max`](https://developer.apple.com/documentation/swift/uint/1539953-max). Perhaps doesn't have the explicit guarantee that it is always greater (it is always `>=`) than any other `Uint` value, but I wonder if the practical guarantee is good enough to simplify a lot of this away.

jsbean:
I could imagine this all relaxing out to:

```Swift
struct Graph <Node: Hashable, EdgeValue: Numeric> {
    struct Edge {
        let value: EdgeValue
        ...
    }
    ...
}
```

```Swift
let graph = Graph<UnassignedNode,UInt>
```

Sources/PitchSpeller/Data Structures/Graph.swift
bwetherfield: yeah it was complaining! I think you may have to do it explicitly when you are using enums with associated values...

Sources/PitchSpeller/Data Structures/Graph.swift
bwetherfield: cool I'll look into Numeric - seems like a good choice

bwetherfield:
I like the generic approach above. Certainly decouples the existing Graph architecture from this specific use case which is nice

bwetherfield:
re UInt.max, I am concerned about integer overflow stuff. That is why coping strategies are hardcoded into all the arithmetic operations I have implemented so far

bwetherfield:
The nice thing about Double.infinity is that it doesn't overflow by design. It just tops out

jsbean:
Yeah, I'm liking that as well! My fetish for encoding invariants into the type system is being soothed.

jsbean:
If you go forth with the generic approach, you'll have to get `EdgeCapacity` to be `Numeric` (which it looks like it's on its already on its way).

I am not entirely thrilled about my choice of removing an `Edge` if it has a value of `0` being baked into the `Graph` implementation. First, we are calling `removeEdge(from:to:)` every time we call `insertEdge`(_:), which is super wasteful. Also, maybe in general cases it's not incorrect to have an `Edge` with a `0` value.

Instead, I feel like we should just filter out all `0`-valued edges before calling `residualNetwork` within `FlowNetwork` (where such a semantic really belongs).

Sources/PitchSpeller/Data Structures/Graph.swift
jsbean: Hmmm, not if the associated value types are themselves `Hashable`.

You can try: 

``` Swift
struct AssociatedType: Hashable {
    let value: Int
}

enum A: Hashable {
    case special
    case normal(AssociatedType)
}
```

Which will compile. If you take away `Hashable` conformance from `AssociatedType`, it will no longer compiler.

bwetherfield:

PR #51, opened 2018-06-27
[PitchSpeller / WIP] Adjacency matrix

bwetherfield:
Conversion from adjacency-list representation of graph to an adjacency-matrix-type representation.
Some benefits include

- faster lookups of edges
- greater potential for 'masking' two graphs with same nodes
- greater consistency with my thesis document


Sources/PitchSpeller/Data Structures/Graph.swift
```diff
@@ -66,32 +71,32 @@ public struct Graph <Value: Hashable>: Hashable {
...
 
-    /// - Returns: All of the `Edge` values in the `Graph`.
+    /// - Returns: All of the `Edge` weights in the `Graph`.
     public var edges: [Edge] {
-        return adjacencyList.flatMap { _, values in values }
+        return Array(adjacencyMatrix.values)
     }
 
     /// - Returns: All of the `Node` values in the `Graph`.
     public var nodes: [Node] {
```
jsbean: I would say get rid of this, and rename `nodeSet` to `nodes: Set<Node>`.

Sources/PitchSpeller/Data Structures/Graph.swift
```diff
@@ -109,69 +114,94 @@ public struct Graph <Value: Hashable>: Hashable {
...
 
     /// Add the given `edge` if it does not currently exist. Otherwise, replaces the edge with the
     /// equivalent `source` and `destination` nodes. If the `value` of the edge is 0, the edge is
     /// removed.
     public mutating func insertEdge(_ edge: Edge) {
-        removeEdge(from: edge.source, to: edge.destination)
-        if edge.value != 0 {
-            adjacencyList.safelyAppend(edge, toArrayWith: edge.source)
-        }
+        adjacencyMatrix[NodePair(edge.source, edge.destination)] = edge
```
jsbean: As discussed pseudo-offline, let's rip `NodePair` out into a generic `struct`:

```Swift
struct Pair <A,B> {
    let a: A
    let b: B
}
```

I might add the initializer:

```Swift
extension Pair {
    init(_ a: A, _ b: B) {
        self.a = a
        self.b = b
    }
}
```

Then, you can extend this to conform conditionally to a bunch of things:

```Swift
extension Pair: Equatable where A: Equatable, B: Equatable { }
extension Pair: Hashable where A: Hashable, B: Hashable { }
```

Then, we can remove the `NodePair` implementation, and replace the calls with `Pair<Node,Node>`.

Sources/PitchSpeller/Data Structures/Graph.swift
```diff
@@ -184,41 +214,52 @@ public struct Graph <Value: Hashable>: Hashable {
...
 
     /// - Returns: A `Graph` with each of the nodes updated by the given `transform`.
-    public func mapNodes <U> (_ transform: (Value) -> U) -> Graph<U> {
-        return Graph<U>(
-            Dictionary(
-                adjacencyList.map { (node,edges) in
-                    (transform(node), edges.map { $0.mapNodes(transform)})
-                }
-            )
+    public func mapNodes <U> (_ transform: (Node) -> U) -> Graph<U, Weight> {
```
jsbean: Looks like you had an argument with the compiler :). I will pull this down to see if I can find the thing the compiler didn't like.

Sources/PitchSpeller/Wetherfield/LookupTable.swift
```diff
@@ -0,0 +1,54 @@
...
+        self.tendency = tendency
+    }
+    
+    init(_ tuple: (Pitch.Class, Tendency)) {
+        self.pitchClass = tuple.0
+        self.tendency = tuple.1
+    }
+}
+
+struct LookupPair: Hashable {
```
jsbean: Ahh, so here we could already use `Pair<LookupIndex, LookupIndex`.

Sources/PitchSpeller/Data Structures/Graph.swift
```diff
@@ -184,41 +214,52 @@ public struct Graph <Value: Hashable>: Hashable {
...
+            { maskAdjacents.contains($0.0) }
+        )
+        self.adjacencyMatrix.merge(
+            mask.adjacencyMatrix,
+            uniquingKeysWith: { $0.mask(against: $1)! }
+        )
     }
 
-    /// - Returns: All of the `Node` values adjacent to the given `node`.
+    /// - Returns: All of the `Node` weights adjacent to the given `node`.
```
jsbean: Here, actually `values` is correct.

Sources/PitchSpeller/Wetherfield/EdgeCapacity.swift
```diff
@@ -0,0 +1,60 @@
...
+enum EdgeCapacity:
+    Comparable
+{
+    case real(Float)
+    case bigM
+}
+
+// MARK: - Comparable
+// bigM edges always 'win' comparisons - otherwise like a regular comparison
+func < (lhs: EdgeCapacity, rhs: EdgeCapacity) -> Bool {
```
jsbean: This might be ever so slightly more elegant:

```Swift
switch (lhs, rhs) {
case let (.real(lhs), .real(rhs)):
    return lhs < rhs
case let (.bigM, .real):
    return false
case let (.real, .bigM):
    return true
}
```

Sources/PitchSpeller/Wetherfield/EdgeCapacity.swift
```diff
@@ -0,0 +1,60 @@
...
+func < (lhs: EdgeCapacity, rhs: Float) -> Bool {
+    return lhs < EdgeCapacity.real(rhs)
+}
+
+func < (lhs: Float, rhs: EdgeCapacity) -> Bool {
+    return EdgeCapacity.real(lhs) < rhs
+}
+
+// MARK: - Math
+func + (augend: EdgeCapacity, addend: EdgeCapacity) -> EdgeCapacity {
```
jsbean: See `<`.

Sources/PitchSpeller/Data Structures/Graph.swift
bwetherfield: I agree with you. The thing I'm struggling with is that I want `nodes` to be a Set internally but an Array publicly... Thoughts on just having `nodes` be a public var Set?

Sources/PitchSpeller/Data Structures/Graph.swift
```diff
@@ -184,41 +214,52 @@ public struct Graph <Value: Hashable>: Hashable {
...
     /// `nil`.
-    public func edgeValue(from source: Node, to destination: Node) -> Double? {
-        guard let edges = adjacencyList[source] else { return nil }
-        for edge in edges {
-            if edge.destination == destination {
-                return edge.value
-            }
-        }
-        return nil
+    public func edgeValue(from source: Node, to destination: Node) -> Weight? {
```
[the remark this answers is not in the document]
bwetherfield: That's the syntactic wizardry I was looking for!

Sources/PitchSpeller/Data Structures/Graph.swift
bwetherfield: Whoops! Yeah, let me know if you can glean anything - thanks :) :)

Sources/PitchSpeller/Wetherfield/EdgeCapacity.swift
bwetherfield: Totally! I may save this change for a different branch / PR... still figuring out what that would be (potentially intedges). As it stands, building this enum on top of `Float` isn't that useful because `Float.infinity` does the same thing as we discussed elsewhere

Sources/PitchSpeller/Data Structures/Graph.swift
jsbean: I would say a `Set` of `nodes` is more correct, and should be the publicly-exposed shape. The `Array` implementation was just what fell out of the adjacency list model.

Sources/PitchSpeller/Data Structures/Graph.swift
jsbean: This works for me:

```Swift
return Graph<U,Weight>(
    nodeSet.map(transform),
    Dictionary(
        adjacencyMatrix.map { nodes, edge in
            return (
                Graph<U,Weight>.NodePair(transform(nodes.source),transform(nodes.destination)),
                edge.mapNodes(transform)
            )
        }
    )
)
```

jsbean: I wonder if we are doing extra work here (mapping over the nodes of `nodeSet`, as well as within each `edge`).

Sources/PitchSpeller/Data Structures/Graph.swift
bwetherfield: Cool - sounds good 👍

Sources/PitchSpeller/Data Structures/Graph.swift
bwetherfield: Yeah, the only reason I have the set of nodes is to cover fully connected nodes, but it does feel a bit redundant. Then again, hard for me to see another way as long as `Node`s are value types. Is this an important function?

jsbean: Ahh, right. Of course. Let's keep this as is.

Sources/PitchSpeller/Data Structures/Graph.swift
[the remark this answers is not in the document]
jsbean: 😎

jsbean:
Would an actual [`Matrix`](https://github.com/dn-m/Structure/blob/master/Sources/DataStructures/Matrix.swift) type be useful here?

bwetherfield:

PR #53, opened 2018-06-27
[PitchSpeller / WIP / Experimental] Edges as pairs

bwetherfield:
What was `NodePair` is replaced by `Edge` which is a type alias for `Pair<Node, Node, Orderedness`.
We also set `typealias Directedness = Orderedness`, which gives us potential for generic graph types, as discussed elsewhere.
TODO: reimplement `Graph` with `adjacencyMatrix: [Edge: Weight]`


jsbean:
I am really liking the direction of this! 

I would highly suggest getting this to compile successfully, though, before progressing further. For two reasons: the compiler will be much more pleasant in the process (it can be a helpful guide, not only a frustrating barrier :)), and further, the regression tests will continuously affirm that you are not flying blind (even if you are crashing into things).

In an ideal world, each commit would build successfully, so that the CI tests can run (of course, this is a "do as I say, not as I do" moment). This way, it makes it convenient to verify that things are working (or not) from the outside (even from mobile).

In the case things _are_ building on your side, we should check if something is weird with CI / pulling from Github.

Sources/PitchSpeller/Data Structures/Graph.swift
```diff
@@ -9,12 +9,16 @@ import Restructure
 import StructureWrapping
 import DataStructures
 
+typealias Directedness = Orderedness
```
jsbean: Is there a mathematical name for this :) ?

Sources/PitchSpeller/Data Structures/Graph.swift
```diff
@@ -304,8 +303,8 @@ extension Graph: CustomStringConvertible {
     public var description: String {
         var result = ""
         var adjacencyList : [Node: [Edge]] = [:]
-        for source in nodeSet {
-            adjacencyList[source] = nodeSet.compactMap{ adjacencyMatrix[NodePair(source, $0)] }
+        for source in nodes {
+            adjacencyList[source] = nodes.compactMap{ adjacencyMatrix[NodePair(source, $0)] }
```
jsbean: Nit: give a little whitespace after `compactMap`.

Sources/PitchSpeller/Data Structures/Pair.swift
```diff
@@ -25,8 +25,39 @@ public struct Pair<A, B, Orderedness> {
...
-extension Pair where Orderedness == Ordered {
+extension Pair {
     public var reversed: Pair<B, A, Orderedness> {
         return Pair<B, A, Orderedness>(b, a)
     }
 }
+
+/// map for elements of Pair
+extension Pair {
+    public func map<T, U> (_ aTransform: (A) -> T, _ bTransform: (B) -> U) -> Pair <T, U, Orderedness> {
```
jsbean: Super nit: I've taken on this style for generics. I could easily be convinced away from this, but so far I like it.

Declaring generic methods, types, etc:

```Swift
func map <T,U> (...) { ... }
func == <T: A, U: B> (...) { ... }
struct Struct <T: A, U: B> { ... }
```

Referencing generic types:

```Swift
func map <T,U> (...) -> Pair<T,U,Orderedness>
```

Synopsis: extra whitespace padding when declaring (because that's where you might constrain the generic types to a constellation of protocols), no whitespace when referencing (because you will not be adding any constraints, and the commas do the job for delimiting).

Sources/PitchSpeller/Data Structures/Pair.swift
```diff
@@ -25,8 +25,39 @@ public struct Pair<A, B, Orderedness> {
...
+    }
+}
+
+public func ==<A, B> (_ lhs: Pair<A, B, Unordered>, _ rhs: Pair<B, A, Unordered>) -> Bool
+    where A: Equatable, B: Equatable {
+    return lhs.a == rhs.b && lhs.b == rhs.a
+}
+
+/// Hashability of Pair where order does not matter
+extension Pair where A: Hashable, A == B, Orderedness == Unordered {
```
jsbean: Very thoughtful!

Sources/PitchSpeller/Data Structures/Pair.swift
```diff
@@ -25,8 +25,39 @@ public struct Pair<A, B, Orderedness> {
...
+
+    func hash(into hasher: inout Hasher) {
```
jsbean: Do you have to define this or only `hashValue`?

Sources/PitchSpeller/Data Structures/Pair.swift
jsbean: Though, you may still need to declare `Hashable` conformance:

```Swift
extension Pair: Hashable where A: Hashable, A == B, Orderedness == Unordered { ... }
```

Might be smart to add a test like so:

```Swift
func testUnorderedPairHashable {
    let set: Set<Pair<Int,Int,Unordered> = [.init(1,2), .init(2,1)]
    XCTAssertEqual(set.count, 1)
}
```

jsbean: If you don't declare the `Hashable` conformance explicitly, the above _shouldn't_ compile. If `hashValue` is implemented very incorrectly, the test should compile and fail.

bwetherfield:

PR #54, opened 2018-06-30
Pairable protocol

bwetherfield:
`Pairable` as a base protocol for Hashable, Equatable, 2-tuple-like data structures (Edges, intervals, etc)
`protocol Symmetric: Pairable` for when the two types are the same.
`protocol Swappable: symmetric` for when the two elements can be flipped meaningfully.

From here we define `OrderedPair`, `UnorderedPair` (Swappable and Symmetric cases respectively) and `Cross` (just Pairable)


Sources/PitchSpeller/Data Structures/Swappable.swift
```diff
@@ -0,0 +1,19 @@
...
+//  Created by Benjamin Wetherfield on 6/30/18.
+//
+
+protocol Swappable: Symmetric {
+    var elput: (A,A) { get }
+    var swapped: Self { get }
+}
+
+extension Swappable {
+    var elput: (A,A) { get { return (tuple.1, tuple.0) } }
```
jsbean: You can forego the `get` here, unless there is also a `set`.

Sources/PitchSpeller/Data Structures/Cross.swift
```diff
@@ -0,0 +1,28 @@
+//
+//  Cross.swift
+//  PitchSpeller
+//
+//  Created by Benjamin Wetherfield on 6/30/18.
+//
+
+struct Cross<T,U>: Pairable {
```
jsbean: Can you add doc comments for everything (`///`)?

 It can be something almost trivial, but even within Xcode, the doc comments are visible when option-cliicking a variable, method, or type. Further, once we get the doc generation machine running again, it will make a huge difference.

I would insert the following `// MARK: - ` comments. See below.

Sources/PitchSpeller/Data Structures/Cross.swift
```diff
@@ -0,0 +1,28 @@
...
+
```
jsbean: `// MARK: - Associated Types

Sources/PitchSpeller/Data Structures/Cross.swift
```diff
@@ -0,0 +1,28 @@
...
+//  PitchSpeller
+//
+//  Created by Benjamin Wetherfield on 6/30/18.
+//
+
+struct Cross<T,U>: Pairable {
+    
+    typealias A = T
+    typealias B = U
+
```
jsbean: `// MARK: - Instance Properties`

Sources/PitchSpeller/Data Structures/Cross.swift
```diff
@@ -0,0 +1,28 @@
...
+//
+
+struct Cross<T,U>: Pairable {
+    
+    typealias A = T
+    typealias B = U
+    
+    let a: T
+    let b: U
+
```
jsbean: `// MARK: - Initializers`

Tests/PitchSpellerTests/OrderedPairTests.swift
```diff
@@ -0,0 +1,54 @@
...
+        let pair = OrderedPair<Int>(3,6)
+        XCTAssertEqual(pair.elput.0, 6)
+        XCTAssertEqual(pair.elput.1, 3)
+    }
+    
+    func testSwapped() {
+        let pair1 = OrderedPair<Int>(3,6)
+        let pair2 = OrderedPair<Int>(6,3)
+        XCTAssertEqual(pair1, pair2.swapped)
+        XCTAssertEqual(pair1, pair1.swapped.swapped)
```
jsbean: Nice.

Tests/PitchSpellerTests/CrossTests.swift
```diff
@@ -0,0 +1,32 @@
...
+//  Created by Benjamin Wetherfield on 7/1/18.
+//
+
+import XCTest
+@testable import PitchSpeller
+
+class CrossTests: XCTestCase {
+    
+    func testEqualityFunction() {
+        let pair1 = Cross<Int, String>(3, "three")
```
jsbean: I could buy the argument that it could be helpful to another human to explicitly declare the generic type members here (`<Int,String>`) — especially in more intricate typing scenarios — but I am relatively certain that the compiler will infer them if you just put: `Cross(3, "three")`.

Might be worth seeing if that's true and trimming these all down. That being said, super micro nit that is not required at all.

Sources/PitchSpeller/Data Structures/Pairable.swift
```diff
@@ -0,0 +1,20 @@
...
+//  Created by Benjamin Wetherfield on 6/30/18.
+//
+
+protocol Pairable {
+    associatedtype A
+    associatedtype B
+    var a: A { get }
+    var b: B { get }
+    var tuple: (A,B) { get }
+    init(_ pair: (A,B))
```
jsbean: Is it critical to demand both of these initializers? If we can get away with it, things may be a bit more agile down the line if only required the `init(_ a: A, _ b: B)` is required, if that.

Sources/PitchSpeller/Data Structures/Pairable.swift
```diff
@@ -0,0 +1,20 @@
...
+//
+//  Created by Benjamin Wetherfield on 6/30/18.
+//
+
+protocol Pairable {
+    associatedtype A
+    associatedtype B
+    var a: A { get }
+    var b: B { get }
+    var tuple: (A,B) { get }
```
jsbean: Is this necessary? I am thinking this could be up to the conforming type how it wants to implement this. Looking down at `Swappable`, I feel that `.swapped` is the only necessary extension made (which could be implemented as `var swapped: Self { return .init(a,b) }`)

Sources/PitchSpeller/Data Structures/Swappable.swift
```diff
@@ -0,0 +1,19 @@
+//
+//  Swappable.swift
+//  PitchSpeller
+//
+//  Created by Benjamin Wetherfield on 6/30/18.
+//
+
+protocol Swappable: Symmetric {
+    var elput: (A,A) { get }
```
jsbean: As above, maybe only `swapped` is required here, and `elput` could be an implementation detail downstream.

Sources/PitchSpeller/Data Structures/Pairable.swift
jsbean: I have generally found that it is nice to only require things which are used within default implementations in protocol extensions. I have over-prescribed before, and it became a bit of a burden downstream.

bwetherfield: Cool. I may keep in `init(_ a: A, _b: A)` for the default definition of `.swapped`...?

Tests/PitchSpellerTests/CrossTests.swift
bwetherfield: Yeah, interesting

PR #55, opened 2018-07-01
Edges from `Pairable`

bwetherfield:
[WIP] Builds on 'Pairable Protocol' PR to implement Edges in terms of Ordered and Unordered pairs.

### DIFF
--- a/Sources/PitchSpeller/Data Structures/DirectedEdge.swift
+++ b/Sources/PitchSpeller/Data Structures/DirectedEdge.swift
@@ -18,8 +18,8 @@ public struct DirectedEdge<Node, Num>: EdgeLike {
         self.weight = weight
     }
     
-    init (_ nodepair: Pair, _ weight: Weight) {
-        self.nodes = nodepair
+    init (_ nodes: Pair, _ weight: Weight) {
+        self.nodes = nodes
         self.weight = weight
     }
 }
--- a/Sources/PitchSpeller/Data Structures/DirectedEdge.swift
+++ b/Sources/PitchSpeller/Data Structures/DirectedEdge.swift
@@ -17,4 +17,9 @@ public struct DirectedEdge<Node, Num>: Weighted & NodePair {
         self.nodes = Pair(source, destination)
         self.weight = weight
     }
+    
+    init (_ nodepair: Pair, _ weight: Weight) {
+        self.nodes = nodepair
+        self.weight = weight
+    }
 }
--- a/Sources/PitchSpeller/Data Structures/DirectedEdge.swift
+++ b/Sources/PitchSpeller/Data Structures/DirectedEdge.swift
@@ -12,4 +12,9 @@ public struct DirectedEdge<Node, Num>: Weighted & NodePair {
 
     var weight: Weight
     let nodes: Pair
+    
+    init (_ source: Node, _ destination: Node, _ weight: Weight) {
+        self.nodes = Pair(source, destination)
+        self.weight = weight
+    }
 }
--- a/Sources/PitchSpeller/Data Structures/DirectedEdge.swift
+++ b/Sources/PitchSpeller/Data Structures/DirectedEdge.swift
@@ -5,7 +5,7 @@
 //  Created by Benjamin Wetherfield on 7/1/18.
 //
 
-public struct DirectedEdge<Node, Num>: Weighted & NodePair {
+public struct DirectedEdge<Node, Num>: EdgeLike {
 
     typealias Weight = Num
     typealias Pair = OrderedPair<Node>
--- a/Sources/PitchSpeller/Data Structures/DirectedEdge.swift
+++ b/Sources/PitchSpeller/Data Structures/DirectedEdge.swift
@@ -10,5 +10,6 @@ public struct DirectedEdge<Node, Num>: Weighted & NodePair {
     typealias Weight = Num
     typealias Pair = OrderedPair<Node>
 
-    
+    var weight: Weight
+    let nodes: Pair
 }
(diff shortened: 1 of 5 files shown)


Sources/PitchSpeller/Data Structures/DirectedEdge.swift
```diff
@@ -0,0 +1,25 @@
+//
+//  DirectedEdge.swift
+//  PitchSpeller
+//
+//  Created by Benjamin Wetherfield on 7/1/18.
+//
+
+public struct DirectedEdge<Node, Num>: EdgeLike {
```
jsbean: I'd prefer no abbrv. in type names (`Num`). Would `Value` be too general?

Would it be helpful to nest this in `Graph`, or do you imagine it would work better outside of that?

Sources/PitchSpeller/Data Structures/EdgeLike.swift
```diff
@@ -0,0 +1,14 @@
+//
+//  EdgeLike.swift
+//  PitchSpeller
+//
+//  Created by Benjamin Wetherfield on 7/1/18.
+//
+
+protocol EdgeLike: Weighted & NodePair {
```
jsbean: Could we get away with just calling this `Edge` if we know that conforming types have to be defined further?

Sources/PitchSpeller/Data Structures/EdgeLike.swift
```diff
@@ -0,0 +1,14 @@
...
+    init(_ nodepair: Pair, _ weight: Weight)
+}
+
+extension EdgeLike where Pair: Swappable {
+    var reversed: Self { return Self(nodes.swapped, weight) }
```
jsbean: Nice.

Sources/PitchSpeller/Data Structures/UndirectedEdge.swift
```diff
@@ -0,0 +1,25 @@
...
+    
+    var weight: Weight
+    let nodes: Pair
+    
+    init (_ source: Node, _ destination: Node, _ weight: Weight) {
+        self.nodes = Pair(source, destination)
+        self.weight = weight
+    }
+    
+    init (_ nodepair: Pair, _ weight: Weight) {
```
jsbean: I could imagine this being named `nodes` or `nodePair`.

Sources/PitchSpeller/Data Structures/UndirectedEdge.swift
```diff
@@ -0,0 +1,25 @@
...
+
+public struct UndirectedEdge<Node, Num>: EdgeLike {
+    
+    typealias Weight = Num
+    typealias Pair = UnorderedPair<Node>
+    
+    var weight: Weight
+    let nodes: Pair
+    
+    init (_ source: Node, _ destination: Node, _ weight: Weight) {
```
jsbean: I don't always do this, but I think I should:

I like to call a higher-level initializer from a lower-level one:

```Swift
init (_ source: Node, _ destination: Node, _ weight: Weight) {
    self.init(Pair(source.destination), weight)
}
```

jsbean: I think this just chisels away at any potential code duplication, therefore potential double-maintainance scenarios.

Sources/PitchSpeller/Data Structures/DirectedEdge.swift
```diff
@@ -0,0 +1,25 @@
+//
+//  DirectedEdge.swift
+//  PitchSpeller
+//
+//  Created by Benjamin Wetherfield on 7/1/18.
+//
+
```
jsbean: As with previous PR, can you add doc comments? These structures are very elegantly wired up, and I think the docs can help show that.

Sources/PitchSpeller/Data Structures/DirectedEdge.swift
bwetherfield: Definitely! That may solve a lot of problems I think in fact

Sources/PitchSpeller/Data Structures/UndirectedEdge.swift
bwetherfield: Yeah, I like `nodes`

Sources/PitchSpeller/Data Structures/UndirectedEdge.swift
bwetherfield: Ah cool!

Sources/PitchSpeller/Data Structures/DirectedEdge.swift
bwetherfield: Is there a good place to put more top-level comments that describe relationships?

Sources/PitchSpeller/Data Structures/EdgeLike.swift
bwetherfield: I hear ya! My issue was with wanting something like:

`public struct Graph<Node: Hashable, Edge: Edge & Hashable>`, but I think the compiler doesn't particularly enjoy that. On reflection, the `EdgeLike` protocol is basically only there to give us the conditional `.reversed` feature, but that's not giving us huge gains, as I can basically just put that in DirectedEdge without much difference...!

bwetherfield: The thing with this `EdgeLike` construction is that it kind of enforces that graphs are weighted too so I think I will do away with it

Sources/PitchSpeller/Data Structures/DirectedEdge.swift
jsbean: If proper types are put in code back-ticks (`Pitch`, `Frequency`), they should become links within the generated docs.

We can put together more top-level places for architectural explanation. We've had versions of that in the past, and would be worth reviving.

PR #68, opened 2018-07-14
[PitchSpeller / WIP] Graph with flags

bwetherfield:
Allows Directedness and Weight flags to be passed into a Graph when declared.

Directedness flag also passes in the Pair type used to represent Edges in the graph as well as the node.

TODO: integrate with existing Graph code.

### DIFF
--- a/Sources/PitchSpeller/Data Structures/_Graph.swift
+++ b/Sources/PitchSpeller/Data Structures/_Graph.swift
@@ -11,3 +11,7 @@ protocol Undirected: Directedness { }
 
 enum WithDirectedEdges: Directed { }
 enum WithUndirectedEdges: Undirected { }
+
+struct _Graph<DirectednessFlag: Directedness> {
+    
+}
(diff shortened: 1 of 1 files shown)


Sources/PitchSpeller/Data Structures/_Graph.swift
```diff
@@ -0,0 +1,72 @@
...
+extension _Graph.Edge where Pair: Directed {
+    
+    // MARK: - Instance Properties
+    
+    var source: _Graph.Node { return nodes.a }
+    var destination: _Graph.Node { return nodes.b }
+}
+
+extension _Graph where Weight: Numeric, Pair: Directed {
+
```
jsbean: Worth shoving a `#warning` in here.

bwetherfield:
@jsbean, might be worth trying to change the base to refactor-graph. It should disappear if it is fully consumed, right?

jsbean:
You be right!

> There are no new commits between base branch 'refactor-graph' and head branch 'graph-with-flags'

Closin'.

PR #69, opened 2018-07-15
[PitchSpeller] Transition from _Graph to Graph

bwetherfield:


Sources/PitchSpeller/Data Structures/_Graph.swift
```diff
@@ -59,8 +59,8 @@ struct _Graph<Weight: Weightedness, Pair: SymmetricPair & Directedness> {
         nodes.append(node)
     }
     
-    func insertEdge(from source: Node, to destination: Node, withWeight weight: Weight) {
-        
+    mutating func insertEdge(from source: Node, to destination: Node, withWeight weight: Weight) {
```
jsbean: I like this naming (`withWeight`) better than what I had before. We should really try to keep this style. I've waffled back and forth, but I think it's time to settle down.

Sources/PitchSpeller/Data Structures/_Graph.swift
```diff
@@ -63,6 +63,10 @@ struct _Graph<Weight: Weightedness, Pair: SymmetricPair & Directedness> {
         edges.append(Edge(source, destination, withWeight: weight))
     }
     
+    func edgeValue (from source: Node, to destination: Node) {
+
```
jsbean: This could be a good opportunity to use a `#warning("TODO")` or something sim.

Sources/PitchSpeller/Data Structures/_Graph.swift
```diff
@@ -65,7 +65,7 @@ struct _Graph<Weight: Weightedness, Pair: SymmetricPair & Directedness & Hashabl
         adjacents[Pair(source, destination)] = weight
     }
     
-    func edgeValue (from source: Node, to destination: Node) -> Weight? {
+    func weight (from source: Node, to destination: Node) -> Weight? {
```
jsbean: Excellent. What I always wanted and should have done from the beginning.

Sources/PitchSpeller/Data Structures/_Graph.swift
```diff
@@ -67,6 +67,7 @@ struct _Graph<Weight: Weightedness, Pair: SymmetricPair & Directedness & Hashabl
     
     mutating func removeEdge(from source: Node, to destination: Node) {
         adjacents[Pair(source, destination)] = nil
+        edges = Array(edges.drop(while: { $0.nodes == Pair(source, destination) }))
```
jsbean: I think you should be able to get away with:

```Swift
edges = Array(edges.drop { $0.nodes == Pair(source, destination) })
```

Sources/PitchSpeller/Data Structures/_Graph.swift
```diff
@@ -51,9 +51,13 @@ struct _Graph<Weight: Weightedness, Pair: SymmetricPair & Directedness & Hashabl
     // MARK: - Instance Properties
      
     var nodes: [Node]
-    var edges: [Edge]
+//    var edges: [Edge]
     var adjacents: [Pair: Weight]
     
+    var edges: [Edge] {
+        return adjacents.compactMap{ Edge($0.key.a, $0.key.b, withWeight: $0.value) }
```
jsbean:
```Swift
return adjacents.compactMap { ... 
```

Sources/PitchSpeller/Data Structures/_Graph.swift
```diff
@@ -60,7 +60,7 @@ struct _Graph<Weight: Weightedness, Pair: SymmetricPair & Directedness & Hashabl
     var adjacents: [Pair: Weight]
     
     var edges: [Edge] {
-        return adjacents.compactMap{ Edge($0.key.a, $0.key.b, withWeight: $0.value) }
+        return adjacents.compactMap{ Edge($0.key, withWeight: $0.value) }
```
jsbean:
```Swift
return adjacents.compactMap { ... 
```

Sources/PitchSpeller/Data Structures/_Graph.swift
```diff
@@ -82,7 +82,14 @@ struct _Graph<Weight: Weightedness, Pair: SymmetricPair & Directedness & Hashabl
     }
     
     func neighbors (of source: Node) -> [Node] {
-        
+        return nodes.compactMap{
```
jsbean: Can this just be:

```Swift
return nodes.compactMap { adjacents[Pair(source,$0)] }
```

jsbean: Also, should neighbors be ordered (`Array`), or would a `Set` make more sense?

jsbean: (Because it looks like your testing to ensure that order doesn't matter, so why not encode that in the type?)

jsbean: Oh whoops, I misread the code a little bit. I'm gonna pull down and toy with that.

jsbean: This works for me:

```Swift
return nodes.compactMap { adjacents.keys.contains(Pair(source,$0)) ? $0 : nil }
```

jsbean: To be slightly more clear:

```Swift
return nodes.compactMap { destination in
    adjacents.keys.contains(Pair(source,destination)) ? destination : nil
}
```

Sources/PitchSpeller/Data Structures/_Graph.swift
```diff
@@ -0,0 +1,136 @@
...
+        init (_ nodes: Pair, withWeight weight: Weight) {
+            self.nodes = nodes
+            self.weight = weight
+        }
+    }
+    
+    // MARK: - Instance Properties
+     
+    var nodes: [Node]
+//    var edges: [Edge]
```
jsbean: Shall we kill this?

Sources/PitchSpeller/Data Structures/_Graph.swift
jsbean: For some reason, my comment on this seemed to disappear. This might be a little cleaner:

```Swift
return nodes.compactMap { destination in
    adjacents.keys.contains(Pair(source,destination)) ? destination : nil
}
```

jsbean: Also, would the result here be better modeled as `Set<Node>` rather than `Array<Node>`?

Sources/PitchSpeller/Data Structures/_Graph.swift
bwetherfield: Yes!

Sources/PitchSpeller/Data Structures/_Graph.swift
```diff
@@ -76,17 +79,20 @@ struct _Graph<Weight: Weightedness, Pair: SymmetricPair & Directedness & Hashabl
...
     
+    /// - Returns: Array of nodes adjacent to `source`
     func neighbors (of source: Node) -> [Node] {
         return nodes.compactMap {
             guard let _ = adjacents[Pair(source, $0)] else { return nil }
             return $0
         }
     }
     
+    /// - Returns: Array of nodes adjacent to `source` out of the supplied set of `nodes`g
```
jsbean: Trailing `g` not as common as a `.`.

bwetherfield: 😆 whoops!

Sources/PitchSpeller/Data Structures/_Graph.swift
```diff
@@ -0,0 +1,293 @@
...
+    }
+    
+    func weight (_ pair: Pair) -> Weight? {
+        return weight(from: pair.a, to: pair.b)
+    }
+    
+    /// - Returns: Array of nodes adjacent to `source`
+    func neighbors (of source: Node) -> [Node] {
+        return nodes.compactMap {
+            guard let _ = adjacents[Pair(source, $0)] else { return nil }
```
jsbean: How about:

```Swift
return nodes.compactMap { node in adjacents[Pair(source,$0)] }
```

jsbean: How about:

```Swift
return nodes.filter { adjacents.keys.contains(Pair(source,$0)) }.map { $0.b }
```

Sources/PitchSpeller/Data Structures/_Graph.swift
```diff
@@ -0,0 +1,293 @@
...
+        }
+    }
+    
+    // MARK: - Instance Properties
+     
+    var nodes: Set<Node>
+    var adjacents: [Pair: Weight]
+    
+    var edges: [Edge] {
+        return adjacents.compactMap { Edge($0.key, withWeight: $0.value) }
```
jsbean: Does this work?

```Swift
return adjacents.compactMap(Edge.init)
```

Sources/PitchSpeller/Data Structures/_Graph.swift
```diff
@@ -0,0 +1,293 @@
...
+            return $0
+        }
+    }
+    
+    func neighbors (of source: Node, from nodes: [Node]) -> [Node] {
+        return neighbors(of: source, from: Set(nodes))
+    }
+    
+    /// - Returns: Array of edges emanating from `source`
+    func edges (from source: Node) -> [Edge] {
```
jsbean: Would  a `Set<Edge>` be a better model for this (i.e., does order matter)?

Sources/PitchSpeller/Data Structures/_Graph.swift
```diff
@@ -0,0 +1,293 @@
...
+    mutating func removeEdge (from source: Node, to destination: Node) {
+        adjacents[Pair(source, destination)] = nil
+    }
+    
+    /// - Returns: Weight of the edge from `source` to `destination` if it exists, else nil
+    func weight (from source: Node, to destination: Node) -> Weight? {
+        return adjacents[Pair(source, destination)]
+    }
+    
+    func weight (_ pair: Pair) -> Weight? {
```
jsbean: Super small nit, and probably another do-as-I-say-not-: higher-level funcs should be placed above the lower level func they call.

Sources/PitchSpeller/Data Structures/_Graph.swift
```diff
@@ -0,0 +1,293 @@
...
+    
+    /// - Returns: Array of nodes adjacent to `source` out of the supplied set of `nodes`.
+    func neighbors (of source: Node, from nodes: Set<Node>) -> [Node] {
+        return nodes.compactMap {
+            guard let _ = adjacents[Pair(source, $0)] else { return nil }
+            return $0
+        }
+    }
+    
+    func neighbors (of source: Node, from nodes: [Node]) -> [Node] {
```
jsbean: Would a Set<Node> be a better model for this (i.e., does order matter)?

Sources/PitchSpeller/Data Structures/_Graph.swift
bwetherfield: I think it would if I added an initializer that accepted (key, value) labeled tuple ?

Sources/PitchSpeller/Data Structures/_Graph.swift
```diff
@@ -0,0 +1,293 @@
...
+        self.nodes = Pair(a, b)
+        self.weight = .unweighted
+    }
+}
+
+extension _Graph where Pair: SwappablePair {
+    
+    // MARK: - Instance Methods
+    
+    mutating func flipEdge (at nodes: Pair) {
```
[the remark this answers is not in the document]
bwetherfield: Yeah fair enough. My mindset is still that they are indexing weights!

Sources/PitchSpeller/Data Structures/_Graph.swift
```diff
@@ -0,0 +1,293 @@
...
+    mutating func insertPath (_ nodes: [_Graph.Node]) {
+        insertPath(Path(nodes))
+    }
+}
+
+extension _Graph where Weight: AsWeight {
+    
+    // MARK: - Instance Methods
+    
+    static func unWeightedVersion (of weightedGraph: _Graph) -> _Graph<WithoutWeights, Pair> {
```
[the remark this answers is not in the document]
bwetherfield: Should work no problem!

Sources/PitchSpeller/Data Structures/_Graph.swift
bwetherfield: Seems good! Is the `lazy` basically a space optimization here?

Sources/PitchSpeller/Data Structures/_Graph.swift
bwetherfield: Yes. I think as soon as the moment arises where this design is more appropriate (e.g. in a test or other comparison operation/ perhaps even a subtracting operation) we should implement.

Sources/PitchSpeller/Data Structures/_Graph.swift
bwetherfield: 👍

Sources/PitchSpeller/Data Structures/_Graph.swift
bwetherfield: Yes, I think you are right. I think for now I have only been using these methods (see also `edges`) as `Sequence` types (for iteration) if I remember rightly. So in that case, I am not sure the increase in complexity is worth the functionality. But I imagine at some point it will come in handy!

Sources/PitchSpeller/Data Structures/_Graph.swift
bwetherfield: Okay, mind blown, it worked!

Sources/PitchSpeller/Data Structures/_Graph.swift
jsbean: That being said, this would work I think:

```Swift
return nodes.filter { adjacents.keys.contains(Pair(source,$0)) }
```

Sources/PitchSpeller/Data Structures/_Graph.swift
jsbean: 👍

PR #79, opened 2018-07-27
Flow network [on Graph refactor]

bwetherfield:


Sources/PitchSpeller/Data Structures/FlowNetwork.swift
```diff
@@ -32,18 +39,78 @@ public struct FlowNetwork <Value: Hashable>: Hashable {
...
+        func findAugmentingPath () -> Bool {
+            guard let path = residualNetwork.shortestUnweightedPath(from: source, to: sink) else {
+                return false
+            }
+            pushFlow(through: path)
+            return true
+        }
+            
+        func pushFlow (through path: UnweightedGraph<Node>.Path) {
+            let minimumEdge = (path.adjacents.compactMap {
```
jsbean: This might be more elegant as:

```Swift
let minimumEdge = (path.adjacents.compactMap(residualNetwork.weight)).min()!
```

Sources/PitchSpeller/Data Structures/FlowNetwork.swift
```diff
@@ -32,18 +39,78 @@ public struct FlowNetwork <Value: Hashable>: Hashable {
...
             residualNetwork.insertPath(path.map { $0 - maximumFlow(of: path) })
         }
         return residualNetwork
     }
+    
+    /// - Returns: (0) The maximum flow of the network and (1) the residual network produced after
+    /// pushing all possible flow from source to sink (while satisfying flow constraints) - with
+    /// saturated edges flipped and all weights removed.
+    var solvedForMaximumFlow: (Double, UnweightedGraph<Node>) {
+        var totalFlow = 0.0
```
jsbean: Is there a way to do this without the state?

Sources/PitchSpeller/Data Structures/FlowNetwork.swift
```diff
@@ -32,18 +39,78 @@ public struct FlowNetwork <Value: Hashable>: Hashable {
...
+    
+    /// - Returns: A minimum cut with nodes included on the `sink` side in case of a
+    /// tiebreak (in- and out- edges saturated).
+    public var minimumCut: (Set<Node>, Set<Node>) {
+        return (sourceSideNodes, notSourceSideNodes)
+    }
+    
+    /// - Returns: Nodes in residual network reachable from the `source`
+    private var sourceSideNodes: Set<Node> {
+        return Set(solvedForMaximumFlow.1.breadthFirstSearch(from: source))
```
jsbean: `solvedForMaximumFlow` could return a labeled tuple, which will make this a little less cryptic.

Sources/PitchSpeller/Data Structures/FlowNetwork.swift
```diff
@@ -32,18 +39,78 @@ public struct FlowNetwork <Value: Hashable>: Hashable {
...
         while let path = residualNetwork.shortestPath(from: source, to: sink) {
             residualNetwork.insertPath(path.map { $0 - maximumFlow(of: path) })
         }
         return residualNetwork
     }
+    
+    /// - Returns: (0) The maximum flow of the network and (1) the residual network produced after
+    /// pushing all possible flow from source to sink (while satisfying flow constraints) - with
+    /// saturated edges flipped and all weights removed.
+    var solvedForMaximumFlow: (Double, UnweightedGraph<Node>) {
```
jsbean: I'd label this tuple: `(flow: Double, graph: UnweightedGraph<Node>)`, or something sim., as usage later becomes cryptic otherwise.

Sources/PitchSpeller/Data Structures/FlowNetwork.swift
```diff
@@ -32,18 +39,78 @@ public struct FlowNetwork <Value: Hashable>: Hashable {
...
+                })
+            }
+        }
+        
+        func addBackEdges () {
+            residualNetwork.edges.map {
+                $0.nodes
+                }.filter {
+                    let weight = residualNetwork.weight($0)!
+                    return weight < 0.001 && weight > -0.001
```
jsbean: Would it make sense to have `epsilon` something that could be injected as a parameter?

Sources/PitchSpeller/Data Structures/FlowNetwork.swift
```diff
@@ -32,18 +39,78 @@ public struct FlowNetwork <Value: Hashable>: Hashable {
...
+            return true
+        }
+            
+        func pushFlow (through path: UnweightedGraph<Node>.Path) {
+            let minimumEdge = (path.adjacents.compactMap {
+                residualNetwork.weight($0)
+                }.min())!
+            totalFlow += minimumEdge
+            path.adjacents.forEach {
+                residualNetwork.updateEdge($0, with: {
```
jsbean:
```
residualNetwork.updateEdge($0, with: { $0 - minimumEdge })
```

Sources/PitchSpeller/Data Structures/FlowNetwork.swift
jsbean: And perhaps we could pull this out of `PerformanceTesting` (and refine it if necessary): [FloatingPoint.isApproximatelyEqual(to:within:)](https://github.com/dn-m/PerformanceTesting/blob/master/Sources/PerformanceTesting/DataSet.swift#L113-L116)

Sources/PitchSpeller/Data Structures/FlowNetwork.swift
bwetherfield: Nice!

Sources/PitchSpeller/Data Structures/FlowNetwork.swift
bwetherfield: Can't think of one. I thought that totalFlow might be useful for testing, but beyond that it shouldn't really be needed...

Sources/PitchSpeller/Data Structures/FlowNetwork.swift
bwetherfield: 👍

Sources/PitchSpeller/Data Structures/FlowNetwork.swift
bwetherfield: 👍

bwetherfield: Nice feature, that

Sources/PitchSpeller/Data Structures/FlowNetwork.swift
bwetherfield: Definitely. I get anxious about equating floats even if it is a comparison against 0.0, which I guess should be an exact value in most implementations (could be off about that!).

Sources/PitchSpeller/Data Structures/FlowNetwork.swift
bwetherfield: Yeah the reason I did the more verbose closure was to avoid ambiguity with two `$0` placeholders meaning different things. First `$0` is a `Pair` and second is a `Weight`

Sources/PitchSpeller/Data Structures/FlowNetwork.swift
```diff
@@ -32,18 +39,78 @@ public struct FlowNetwork <Value: Hashable>: Hashable {
...
+                    minuend in
+                    minuend - minimumEdge
+                })
+            }
+        }
+        
+        func addBackEdges () {
+            residualNetwork.edges.map {
```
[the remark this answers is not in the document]
bwetherfield: Word, totally

Sources/PitchSpeller/Data Structures/FlowNetwork.swift
jsbean: Ahh, that's fair. Maybe then I'd go full verbose:

```Swift
path.adjacents.forEach { edge in 
    residualNetwork.updateEdge(edge, with: { value in value - minimumEdge })
}
```

Sources/PitchSpeller/Data Structures/FlowNetwork.swift
jsbean: Fair enough!

jsbean: It might be worth it to do some tests around that. I feel like it _should_ be safe to compare against `0.0`.

Sources/PitchSpeller/Data Structures/FlowNetwork.swift
```diff
@@ -41,9 +41,7 @@ public struct FlowNetwork <Node: Hashable> {
         }
             
         func pushFlow (through path: UnweightedGraph<Node>.Path) {
-            let minimumEdge = (path.adjacents.compactMap {
-                residualNetwork.weight($0)
-                }.min())!
+            let minimumEdge = (path.adjacents.compactMap(residualNetwork.weight).min())!
```
jsbean: Does this whole thing need to be wrapped in parens, or can the forced optional unwrapping happen only at the `min()` call, e.g.:

```Swift
let minimumEdge = path.adjacents.compactMap(residualNetwork.weight).min()!
```

As the compact map will return an `[Weight]` (non-optional), where the `min()` call will return `nil` if this array is empty.

Which brings up an interesting question: do we have guarantees somewhere that a `Path` (and therefore `path.adjacents`) will always be non-empty? That could be worth looking into.

Sources/PitchSpeller/Data Structures/_Graph.swift
```diff
@@ -156,6 +156,15 @@ struct _Graph<Weight: Weightedness, Pair: SymmetricPair & Directedness & Hashabl
         self.nodes = nodes
         self.adjacents = adjacents
     }
+
+    // Initializer from phased-out Graph type
+    init (_ graph: __Graph<Node>) {
+        nodes = Set(graph.nodes)
+        adjacents = [:]
+        for edge in graph.edges {
+            adjacents[Pair(edge.source, edge.destination)] = (edge.value as! Weight)
```
jsbean: I'm not super convinced by the need to forcibly cast here (`as!`). Though, is this just a temporary adapter to the deprecated `__Graph`?

jsbean: Could this be nested in an extension as:

```Swift
extension _Graph where Weightedness == Double {
init (_ graph: __Graph<Node>) {
         nodes = Set(graph.nodes)
         adjacents = [:]
         for edge in graph.edges {
             adjacents[Pair(edge.source, edge.destination)] = edge.value
        }
    }
}
```

Sources/PitchSpeller/Data Structures/__Graph.swift
```diff
@@ -8,9 +8,7 @@
 import DataStructures
 
 /// Minimal implementeation of a Directed Graph with Weighted (/ Capacious) Edges.
-public struct Graph <Value: Hashable>: Hashable {
-
-    public typealias Node = Value
+public struct __Graph <Node: Hashable>: Hashable {
```
jsbean: Are we nearing a point where we can remove this?

PR #80, opened 2018-07-27
Fix assignNodes variable in terms of new minimumCut

bwetherfield:
Fixes corner case involving nodes that are reachable from neither the source nor the sink in the residual network.


jsbean:
This looks great! Want me to merge it in?

PR #103, opened 2018-09-07
[NFC] Make tiny improvements to implementation of Graph

jsbean:


jsbean:
@bwetherfield, what's the deal with `addBackEdges()`. Is it not necessary, because it wasn't being used.

Is there a test the could be written to assert the necessity (or lack thereof) of its existence?

jsbean:
@bwetherfield, objections to merging this in?

bwetherfield:
No objections! Yes - back edges need to be added during the algorithm rather than in one fell swoop so it was phased out.

PR #109, opened 2018-09-22
Push down Graph to dn-m/Structure/DataStructures

jsbean:
This is a work-in-progress PR which refactors the previous design (a single, generic `Graph` struct) into a design using a constellation of protocols.

Ultimately, the protocol universe will look like this:

```Swift
protocol Graph {
    associatedtype Node: Hashable
    associatedtype Weight
    associatedtype Edge: SymmetricPair & Hashable where Edge.A == Node
    var nodes: Set<Node> { get set }
    var adjacents: [Edge: Weight] { get set }
    init(_ nodes: Set<Node>, _ adjacents: [Edge: Weight])
}
struct Unweighted { }
protocol DirectedGraph: Graph where Edge == OrderedPair<Node> { }
protocol UndirectedGraph: Graph where Edge == UnorderedPair<Node> { }
protocol WeightedGraph: Graph where Weight: Numeric { }
protocol UnweightedGraph: Graph where Weight == Unweighted { }
```

The concrete types will be:

```Swift
struct WeightedDirectedGraph <Node: Hashable, Weight: Numeric>:  WeightedGraph, DirectedGraph { }
struct WeightedUndirectedGraph <Node: Hashable, Weight: Numeric>: WeightedGraph, UndirectedGraph { }
struct UnweightedDirectedGraph <Node: Hashable>: UnweightedGraph, DirectedGraph { }
struct UnweightedUndirectedGraph <Node: Hashable>: UnweightedGraph, UndirectedGraph { }
```

We can carefully place each bit of functionality into the correct sub-protocol of `Graph` (e.g., only expose `Weight`-related things to `WeightedGraph`-conforming types, etc.).

@bwetherfield, I could imagine bike-shedding the naming a little. For example, I could imagine the protocols having the `Protocol` suffix, then making the concrete types more concise: 

```Swift
// Instead of `UnweightedDirectedGraph<Node>`
struct DirectedGraph <Node>: UnweightedGraphProtocol, DirectedGraphProtocol { }
// Instead of `UnweightedUndirectedGraph<Node>`
struct Graph <Node>: UnweightedGraphProtocol, UndirectedGraphProtocol { }
```

That way, the call site will be cleaner, but the protocol constellation more wordy.

Currently, all of these types are underscored (e.g., `_UnweightedGraph`, etc.)


jsbean:
@bwetherfield, I am now thinking we can completely get rid of `Weight` from the `Graph` and `UnweightedGraph` protocols. Stand by.

jsbean:
The way that `Path` is currently implemented, it is basically just a `Graph`. Should we just get ride of that and introduce an `insert(_ graph: Graph)` which does the same thing?

One thing to consider is merge conflicts (if you insert a `Graph` with node `A` into another `Graph` which also have a node `A`). Hmm.

jsbean:
The current state of this PR now defines these types (omitting some associated type constraints for clarity):

```Swift
protocol GraphProtocol { }
protocol DirectedGraphProtocol: GraphProtocol { }
protocol UndirectedGraphProtocol: GraphProtocol { }
protocol WeightedGraphProtocol: GraphProtocol { }
protocol UnweightedGraphProtocol: GraphProtocol { }

struct Graph <Node>: UnweightedGraphProtocol, UndirectedGraphProtocol { }
struct DirectedGraph <Node>: UnweightedGraphProtocol, DirectedGraphProtocol { }
struct WeightedGraph <Node, Weight>: WeightedGraphProtocol, UndirectedGraphProtocol { }
struct WeightedDirectedGraph <Node, Weight>: WeightedGraphProtocol, DirectedGraphProtocol { }
```

bwetherfield:
> One thing to consider is merge conflicts (if you insert a `Graph` with node `A` into another `Graph` which also have a node `A`). We could provide a similar `uniquingKeys` closure à la `Dictionary`…

With `nodes` as a `Set<Node>`, we should be okay I think, unless I am missing things. But `uniquingKeys` would have to come in for edge weights as I see it (where the pair of connecting nodes is the same)... basically anywhere that the underlying data structure is a `Dictionary`.

jsbean:
Ahh, yes. I misspoke (read: I misunderstood my own problem). The issue (as you stated) is only for `WeightedGraphProtocol`-conforming types, wherein the underlying data structure is `[Edge: Weight]`.

Thus, we could implement a `+` / `insert(contentsOf:)` function for `UnweightedGraphProtocol`-conforming types where the underlying data structure is a `Set<Edge>`. For `WeightedGraphProtocol`-conforming types, we could implement `insert(contentsOf:uniquingEdgesWith:)`.

This brings up an interesting issue. It would be nice to add `Additive` monoid conformance to all of the `Graph` types (by defining an `identity` (`empty`) element, and a `+` operation). The `+` operator usually (always?) _implies_ commutativity, though in the context of a monoid it doesn't necessarily mean so.

Thus, unweighted graph additions are commutative (and thus are commutative monoids under addition), yet weighted graph additions are not commutative (and thus are merely monoids under addition) for our design. Of course, in our design, there cannot be multiple edges between a given pair of nodes.

In any case, there is a lot of headroom for wheel-spinning here which I will now step back from.

jsbean:
I really wanted the parameterized phantom-typed `Graph` built on a single generic struct to work, but as we ran into the multiple-conformance-to-the-same-protocol issue, this seemed shakier. Especially re: extensibility.

In the end, it seemed likely that we would just provide typealiases for the common settings (`WeightedGraph<Node,Weight> = ...`). So ultimately, they might as well just be in their own types.

jsbean:
For future reference, all of this has led to the `Graph` universe being pushed down into [dn-m/Structure/DataStructures](https://github.com/dn-m/Structure/tree/master/Sources/DataStructures/Graph)

PR #113, opened 2018-09-30
Flow tests

bwetherfield:
PR to add more tests to `FlowNetwork`!

For now the issue of `source` and `sink` being in `FlowNetwork` has been put in a test that will ultimately be made to fail and rewritten.


jsbean:
Very nice. This makes me want to label the tuple result of `minimumCut` for sure!

PR #114, opened 2018-09-30
Throw in flow network

bwetherfield:
The PR rehashes the Edmunds-Karp implementation to make it more throw-catch-centric.


Sources/SpelledPitch/PitchSpeller/Wetherfield/FlowNetwork.swift
```diff
@@ -35,44 +35,55 @@ extension FlowNetwork {
...
+            throw EdgeIs.saturated
+        } else {
+            updateEdge(edge) { weight in weight - amount }
         }
     }
 
-    /// Inserts an edge in the opposite direction of the given `edge` with the minimum flow
-    mutating func updateBackEdge(_ edge: Edge, by minimumFlow: Weight) {
+    /// Inserts an edge in the opposite direction of the given `edge` with the minimum capacity
+    mutating func enableBackFlow(through edge: Edge, equalTo amount: Weight) {
```
jsbean: I remember not liking this signature very much before. Would this be better:

```Swift
func enableBackFlow(_ flow: Weight, through edge: Edge) { ... } 
```

Sources/SpelledPitch/PitchSpeller/Wetherfield/FlowNetwork.swift
```diff
@@ -35,44 +35,55 @@ extension FlowNetwork {
...
+    enum EdgeIs: Error {
+        case absent
+        case saturated
+        case oversaturated
+    }
+    
+    /// Increases the flow through the given `edge` by `amount`. Throws `edgeAbsent` if `edge`
+    /// is not found in the network's underlying graph, and `capacityExceeded` if the amount
+    /// being added exceeds the remaining capacity of the edge.
+    mutating func increaseFlow(through edge: Edge, by amount: Weight) throws {
```
jsbean: An alternative to throwing here could be to return an enum with the status of the `Edge`:

```Swift
func increaseFlow(through edge: Edge, by amount: Weight) -> EdgeIs { ... }
```

Sources/SpelledPitch/PitchSpeller/Wetherfield/FlowNetwork.swift
```diff
@@ -35,44 +35,55 @@ extension FlowNetwork {
...
-    /// the `edge` is now `0`, removes the `edge` from the network. Updates the reverse of the given
-    /// `edge` by the given `minimumFlow`.
-    mutating func pushFlow(through edge: Edge, by minimumFlow: Weight) {
-        reduceFlow(through: edge, by: minimumFlow)
-        removeEdgeIfFlowless(edge)
-        updateBackEdge(edge, by: minimumFlow)
+    /// Increases the flow through the given `edge` by the given `minimumCapacity`. If `edge` is
+    /// saturated (its flow equals its capacity), removes the `edge` from the network. Increases
+    /// the capacity for flow in the opposite direction (reversed edge) by `minimumCapacity`.
+    mutating func pushFlow(through edge: Edge, equalTo amount: Weight) {
```
jsbean: If instead of throwing (see above), you could switch directly over the result of `increaseFlow(through:by:)`:

```Swift
switch increaseFlow(through: edge, by: amount) {
    case .isSaturated:
        ...
    ...
}
```

Sources/SpelledPitch/PitchSpeller/Wetherfield/FlowNetwork.swift
jsbean: It might be nice to decouple the `.absent` case from the saturation cases. For example, if edge absent-ness was an actual throwable error, while the saturation state could be encoded into an `enum` which is handled as a return value.

The `.absent` case seems much more general (belonging to `Graph`, rather than `FlowNetwork`), while saturation is a `FlowNetwork`-specific attribute.

Sources/SpelledPitch/PitchSpeller/Wetherfield/FlowNetwork.swift
bwetherfield: Yes! I was torn, and wanted to keep to a loose "edge-first" pattern we seem to have going in signatures. OK to divert from that?

Sources/SpelledPitch/PitchSpeller/Wetherfield/FlowNetwork.swift
bwetherfield: This is elegant. Like it.

Sources/SpelledPitch/PitchSpeller/Wetherfield/FlowNetwork.swift
bwetherfield: Definitely!

PR #118, opened 2018-10-12
Refactor neutralLetterName calculation

bwetherfield:
Tried to rewrite the `static func neutral(for pitchClass: Pitch.Class) -> LetterName?` function in terms closer to its comment:
```swift
/// - Returns: The `LetterName` which corresponds to the `.neutral` `ModifierDirection` for the
    /// the given `pitchClass`, if such a `LetterName` exists. Otherwise, `nil`.
    static func neutral(for pitchClass: Pitch.Class) -> LetterName? {
        guard let category = Pitch.Spelling<EDO12>.Category
            .category(for: pitchClass) else { return nil }
        guard let modifier = category.modifiers[.neutral] else { return nil }
        switch modifier {
        case .natural:
            return LetterName.default(for: pitchClass)
        case .flat:
            return LetterName.default(for: pitchClass + 1)
        case .sharp:
            return LetterName.default(for: pitchClass - 1)
        }
    }
```
The compiler has made me declare an explicit `TuningSystem`, which may be a good point. How does all this change for other tuning systems and how do we account for that, @jsbean? On the other hand, I was thinking that the above logic might work for all `EDO\(n)` tuning systems where `n` is a multiple of 12. It doesn't seem like I can instantiate `Pitch.Spelling` with a protocol as its generic argument... a quandary.

Can the above function somehow exist in `Pitch.Spelling where TuningSystem: EDO`? The problem is that a specific constructor for `LetterName` is supposed to call `neutral`:
```swift
extension LetterName {

    /// Create a `LetterName` with the given `pitchClass` and `modifierDirection`.
    init?(pitchClass: Pitch.Class, modifierDirection: ModifierDirection) {
        if pitchClass == 8 {
            switch modifierDirection {
            case .down: self = .a
            case .up: self = .g
            default: return nil
            }
            return
        }
        guard
            let neutral = LetterName.neutral(for: pitchClass),
            let adjusted = neutral.adjusted(for: pitchClass, with: modifierDirection)
        else {
            return nil
        }
        self = adjusted
    }
}
```
But perhaps this behavior can be achieved with
```swift
extension Pitch.Spelling where TuningSystem: EDO {
    static func letterName (pitchClass: Pitch.Class, modifierDirection: ModifierDirection) 
                                            -> LetterName? {
...
    }
}

### CODE REVIEW

File: Sources/SpelledPitch/PitchSpeller/Wetherfield/Pitch.Spelling.Category.swift [hunk 1]
```diff
@@ -119,22 +119,18 @@ extension LetterName {
...
-        case 3,10:
-            return LetterName.default(for: pitchClass + 1)!
-        // `Pitch.Class` values which can be spelled with a `natural` modifier need no adjustment
-        case 0...11:
-            return LetterName.default(for: pitchClass)!
-        default:
-            return nil
+        guard let category = Pitch.Spelling<EDO12>.Category
+            .category(for: pitchClass) else { return nil }
+        guard let modifier = category.modifiers[.neutral] else { return nil }
```
jsbean: :+1:

jsbean:
This is very compelling: 

```Swift
extension Pitch.Spelling where TuningSystem: EDO {
    static func letterName(pitchClass: Pitch.Class, modifierDirection: ModifierDirection) -> LetterName?
    {
        ...
    }
}
```

Do you want to give that a shot and see how it ripples downstream?

jsbean:
@bwetherfield, looks very nice. See #122 for updated tests.

jsbean:
Re: `TuningSystem`. I'm feeling a little uneasy about having injected this into `Pitch.Spelling` (and downstream into `SpelledPitch`, etc.). I am going to start an issue to get your feedback on this.

bwetherfield:
#124

jsbean:
@bwetherfield, is it OK to wait on this for a sec while we look into #127? I think this will be simpler to add on after that (merge-conflict-wise).

PR #123, opened 2018-10-16
Diversify indexing types in PitchSpeller

bwetherfield:
Where currently we have two graphs in PitchSpeller, both of which are indexed by `Int` values, this PR proposes wrapping the conversion between the two indices in concrete types, so that the larger graph uses `Cross<Int, Tendency>` indices cutting down on the need for encoding/decoding operations over integers.


Sources/SpelledPitch/PitchSpeller/Wetherfield/Wetherfield.swift
```diff
@@ -68,23 +74,38 @@ struct PitchSpeller {
...
+            sinkSide.remove(flowNetwork.sink)
             let downNodes = sourceSide.map { index in AssignedNode(index, .down) }
             let upNodes = sinkSide.map { index in AssignedNode(index, .up) }
             return downNodes + upNodes
         }
 
         return assignedNodes
-            .sorted { $0.index < $1.index }
-            .dropFirst(2)
+            .sorted {
```
jsbean: Would this be semantically equivalent?

```Swift
.sorted {
    if $0.index.a < $1.index.a { return true }
    return $0.index.a == $1.index.a && UInt8($0.index.b.rawValue) < UInt8($1.index.b.rawValue)
}
```

Sources/SpelledPitch/PitchSpeller/Wetherfield/Wetherfield.swift
```diff
@@ -68,23 +74,38 @@ struct PitchSpeller {
...
+                switch (
+                    $0.index.a < $1.index.a,
+                    $0.index.a == $1.index.a,
+                    UInt8($0.index.b.rawValue) < UInt8($1.index.b.rawValue)
```
jsbean: My feeling is that we shouldn't have to convert this to a `UInt8` and extract out the `rawValue`. This comparison seems like it may be better suited as part of the type (`Tendency`, unless I'm mistake) itself.

Sources/SpelledPitch/PitchSpeller/Wetherfield/Wetherfield.swift
jsbean: Overall, my intuition is that this comparison should be in the conformance of `AssignedNode: Comparable`. I'm gonna pull down to poke at it.

jsbean: See #125.

Sources/SpelledPitch/PitchSpeller/Wetherfield/Wetherfield.swift
bwetherfield: 💯This was a case of saying "whatever you say" to the compiler ! I am hoping to deprecate this with #126

jsbean:
How's this feeling @bwetherfield ? I could imagine getting this into `master`, piling on the other PRs in after.

PR #125, opened 2018-10-16
Make AssignedNode Comparable

jsbean:
This PR makes `AssignedNode` conform to the `Comparable` protocol (simplifying comparison for sorting down the line).


bwetherfield:
What do you think about adding in a layer of 
```swift
extension Cross: Comparable where A: Comparable & Equatable, B: Comparable {
    func < (lhs: Cross<A,B>, rhs: Cross<A,B>) {
        return lhs.a < rhs.a || lhs.a == rhs.a && lhs.b < rhs.b
    }
}
```
and then potentially adding

```swift
extension PitchSpellerNode: Comparable where PitchSpellerNode.index.Type: Comparable {
    func < (lhs: PitchSpellerNode, rhs: PitchSpellerNode) {
        return lhs.index < rhs.index
    }
}
```

jsbean:
I was wondering the exact same thing. However, I was nervous adding the implicit ordering of parameters in `Cross` (that `.a` is checked for comparability before `.b`).

@bwetherfield, I could be convinced not to feel nervous about that, though.

bwetherfield:
This could be encoded in types, e.g. `LeftWeightedCross`, `RightWeightedCross` and boring old `Cross`. Thoughts?

bwetherfield:
I've been thinking about this a bit more. I think it's fine for `.a` to be compared before `.b`, since if you want the other behavior you can use `Cross<Y,X>` instead of `Cross<X,Y>`. I think getting that lexicographical ordering for free seems fairly natural to me. What do you think, @jsbean?

jsbean:
That works for me. I would say to put a clear description in the `<` implementation, though!

PR #126, opened 2018-10-16
Improve runtime of spell function with dictionaries

bwetherfield:
Using a dictionary to store pitches eliminates the need to sort arrays (an O(nlog n) cost) and sets stage for `EventID` type indexing (in dictionaries), which could be more flexible / stable going forward.


Sources/SpelledPitch/PitchSpeller/Wetherfield/Wetherfield.swift
```diff
@@ -65,7 +65,7 @@ struct PitchSpeller {
     // MARK: - Initializers
 
     /// Create a `PitchSpeller` to spell the given `pitches`, with the given `parsimonyPivot`.
-    init(pitches: [Pitch], parsimonyPivot: Pitch.Spelling<EDO12> = .init(.d)) {
+    init(pitches: [Int: Pitch], parsimonyPivot: Pitch.Spelling<EDO12> = .init(.d)) {
```
jsbean: Should the caller need to supply the index here (as part of the API), or should that be done as part of an implementation detail. E.g.,

```Swift
init(pitches: [Pitch], ...) {
    self.pitches = Dictionary(pitches.enumerated())
    ...
}
```

* not verified, but something like that?

jsbean: In a sense, I don't think the user of `Pitch.Speller` should even know why such indices would exist.

jsbean: Ahh, jeez, just started a review by accident.

bwetherfield: I figure that this `Int` may one day become a more general unique EventID tied to something in the score model, so you would not necessarily have `0,1,2,3,...`, and so I do want to explicitly tie each `Pitch` value to its index.

Sources/SpelledPitch/PitchSpeller/Wetherfield/Wetherfield.swift
```diff
@@ -90,30 +90,26 @@ struct PitchSpeller {
...
-        let pitch = self.pitch(node: up.index)
+        let pitch = self.pitch(node: up.index)!
         let tendencies = TendencyPair((up.assignment, down.assignment))
         let spelling = Pitch.Spelling<EDO12>(pitchClass: pitch.class, tendencies: tendencies)!
         return SpelledPitch(try! pitch.spelled(with: Pitch.Spelling<EDO48>(spelling)))
     }
 
     /// - Returns: The `Pitch` value for the given `node` value.
-    private func pitch(node: Cross<Int, Tendency>) -> Pitch {
+    private func pitch(node: Cross<Int, Tendency>) -> Pitch? {
```
jsbean: If this is `private`, maybe we could just return a `Pitch` (non-`Optional`), as that would perhaps be a programmer error, and something that should never even be possible. Thus you could:

```Swift
return pitches[node.a]!
```

Sources/SpelledPitch/PitchSpeller/Wetherfield/Wetherfield.swift
jsbean: Ahh, interesting.

Perhaps there could be a context-less API for the speller. And then a progressively-disclosed increasingly-contextual set of APIs (e.g., to handle adjacency, segmentation, etc.).

jsbean: Whoops, was that PR description there the whole time 🤦‍♂️

bwetherfield: I do think the PitchSpeller will need to be passed adjacency information in addition to raw pitches as the same set of pitches can get spelled differently depending on how close they are to each other and in what order they occur (basically adjacency info)

bwetherfield: I am thinking of the `Int` as the `Node` values of the adjacency graph that get passed in

bwetherfield: ... and which also contribute to the `typealias NewNode = Cross<Node, Tendency>` node values in the `flowNetwork`

PR #131, opened 2018-10-22
Let Flow Network conform to Graph protocols directly

bwetherfield:
Leverages the `...GraphProtocol` ecosystem to remove the object composition of a graph within `FlowNetwork`. All graph type methods and storage now belong to the flow network itself.


jsbean:
See #141.

Issue #149, opened 2018-11-02
[PitchSpeller] Actually use parsimonyPivot


jsbean:
cc @bwetherfield.

Currently, `parsimonyPivot` is unused within `PitchSpeller.spell()`. It doesn't make it out of `Pitch.Speller.init(_: [Int: Pitch], parsimonyPivot: Pitch.Spelling)`.

What is required to hook this up?

jsbean:
Closed by #151.

PR #151, opened 2018-11-03
Refactor flow network indices as casewise sink, source or internal

bwetherfield:
This PR solves #149.

The flow network's node indices are nested in a monad-like `enum` structure with cases `.source`, `.sink` and ``.`internal`(Cross<Int,Tendency>)``. Where `Cross<Int,Tendency>` is what was previously the index for `PitchSpellingNode` values.

 `pitch` is a map from the new index type to `Pitch?`, which functions like the old dictionary of pitches but is initialized as follows.

```swift
init(pitches: [Int: Pitch], parsimonyPivot: Pitch.Spelling = .init(.d)) {
        self.pitch = { index in
            switch index {
            case .source, .sink:
                return Pitch(value: parsimonyPivot.pitchClass.value)
            case .`internal`(let cross):
                return pitches[cross.a]
            }
        }
        self.parsimonyPivot = parsimonyPivot
        self.pitchNodes = PitchSpeller.internalNodes(pitches: pitches)
        self.flowNetwork = FlowNetwork(
            source: .source,
            sink: .sink,
            internalNodes: pitchNodes
        )
```

Because `.source` and `.sink` have their `Pitch` values encoded in the `pitch` function type, the proper pitch of the parsimony pivot can be incorporated into the algorithm. 

Ultimately I think this `enum` type could and should be lifted into `FlowNetwork`...


Sources/SpelledPitch/PitchSpeller/Wetherfield/Wetherfield.swift
```diff
@@ -28,8 +34,14 @@ struct PitchSpeller {
     )
 
     static let weightCarrying = WeightCarrying.build(from: PitchSpeller.tendencyGraph)
-    static let tendencyMask: WeightCarrying<WeightedGraph<Cross<Int,Tendency>,Double>> = weightCarrying.pullback { $0.b }
-
+    static let tendencyMask: WeightCarrying<WeightedGraph<FlowNode<Cross<Int,Tendency>>,Double>> = weightCarrying.pullback {
+        switch $0 {
```
jsbean: When switching over something, might as well give it a name, like `node` or something.

jsbean: `.pullback { node in switch node { ... } } ...`

Sources/SpelledPitch/PitchSpeller/Wetherfield/Wetherfield.swift
```diff
@@ -28,8 +34,14 @@ struct PitchSpeller {
...
+        case .source: return .down
+        case .sink: return .up
+        case .`internal`(let index): return index.b
```
jsbean: Do you have to backick this each time, or only the declaration in the enum. I think because its a proper case, it should be fine?

Sources/SpelledPitch/PitchSpeller/Wetherfield/Wetherfield.swift
```diff
@@ -42,19 +54,28 @@ struct PitchSpeller {
...
+    struct InternalAssignedNode {
+        let index: Cross<Int, Tendency>
+        let assignment: Tendency
+        init(_ index: Cross<Int, Tendency>, _ assignment: Tendency) {
+            self.index = index
+            self.assignment = assignment
+        }
     }
 
+//    /// - Returns: The nodes for the `Pitch` at the given `index`.
```
jsbean: Should we burn this?

Sources/SpelledPitch/PitchSpeller/Wetherfield/Wetherfield.swift
```diff
@@ -8,8 +8,14 @@
 import DataStructures
 import Pitch
 
+enum FlowNode<Index>: Hashable where Index: Hashable {
```
jsbean: I really like this, as this cleans some computation up downstream where we had to awkwardly filter it out.

Sources/SpelledPitch/PitchSpeller/Wetherfield/Wetherfield.swift
```diff
@@ -143,11 +172,11 @@ extension FlowNetwork where Node == PitchSpellingNode.Index, Weight == Double {
     }
 }
 
-extension PitchSpeller.AssignedNode: Comparable {
-    static func < (lhs: PitchSpeller.AssignedNode, rhs: PitchSpeller.AssignedNode) -> Bool {
-        return lhs.index < rhs.index
-    }
-}
+//extension PitchSpeller.AssignedNode: Comparable {
```
jsbean: Shall we burn this?

Tests/SpelledPitchTests/PitchSpellerTests.swift
```diff
@@ -13,12 +13,12 @@ import Pitch
 class PitchSpellerTests: XCTestCase {
 
     // MARK: - AssignedNode
-
-    func testAssignedNodeComparable() {
-        let a = PitchSpeller.AssignedNode(Cross(0,.down), .down)
-        let b = PitchSpeller.AssignedNode(Cross(1,.up), .down)
-        XCTAssertLessThan(a,b)
-    }
+//
```
jsbean: Should this be replaced with `FlowNode` comparison or sim.?

Tests/SpelledPitchTests/PitchSpellerTests.swift
```diff
@@ -13,12 +13,12 @@ import Pitch
 class PitchSpellerTests: XCTestCase {
```
jsbean: Can we add a `testSpellEightOverCNatural()` which verifies that we get an `(.a, .flat)` result?

Sources/SpelledPitch/PitchSpeller/Wetherfield/Wetherfield.swift
bwetherfield: Ah, okay

Tests/SpelledPitchTests/PitchSpellerTests.swift
bwetherfield: Not sure that this comparable conformance is needed in the context anymore, because we are no longer sorting `Node`s by value. Thoughts?

Tests/SpelledPitchTests/PitchSpellerTests.swift
bwetherfield: Yes, that's definitely the plan! I think there will be a bit more organizing needed before we can get a green light on that, but I can always throw in a failing test for now of course / cheat for now!

Sources/SpelledPitch/PitchSpeller/Wetherfield/Wetherfield.swift
bwetherfield: Maybe a "burn" PR after we get through this PR flurry or do you prefer doing it along the way?

jsbean: I'd say lets clear out any dead / commented-out code before merging.

Tests/SpelledPitchTests/PitchSpellerTests.swift
jsbean: Ahh, then all good!

Tests/SpelledPitchTests/PitchSpellerTests.swift
jsbean: How about: let's merge this without adding the test. But then, add a PR where the first commit is such a failing test. Then we can work to green-light it.

PR #153, opened 2018-11-03
Switch out WeightCarrying for AdjacencyCarrying...

bwetherfield:
where the weights used are just `1` and hence standing in for adjacency information.


Sources/SpelledPitch/PitchSpeller/Wetherfield/Wetherfield.swift
```diff
@@ -16,19 +16,20 @@ protocol PitchSpellingNode: Hashable {
 struct PitchSpeller {
 
     // FIXME: Flesh out for all tendencies
-    static let tendencyGraph: WeightedGraph<Tendency,Double> = WeightedGraph(
+    static let tendencyGraph: Graph<Tendency> = Graph(
```
jsbean: :+1:

jsbean: In the fullness of time, I think we should move out all of the masking-related stuff to an extension, but no rush.

PR #163, opened 2018-11-05
Spell self as pivot correctly

bwetherfield:
Implements `() -> .e` with `parsimonyPivot = .e` and so on...


bwetherfield:
The test I have added passes with a hack. I think it would be nice to add an initializer for `Pitch` of the form: 
```swift
extension Pitch {
    init (_ pitchClass: Pitch.Class, _ octave: Int) {
        pitchClass + ((octave + 1) * 12)
    }
}
```
not sure if the implementation is right!

PR #165, opened 2018-11-05
[NFC] Reorganize PitchSpeller struct into extensions

jsbean:
This PR organizes the `PitchSpeller` struct into extensions containing initializers, type properties, instance properties, and instance methods.

It also moves a few type properties out to private free functions.


bwetherfield:
Yay!!!

PR #175, opened 2018-11-07
Big M Implementation

bwetherfield:
To be pulled into `dyad-tests` when we first get red flags. Should choose either F#-A# _or_ Gb - Bb over `.d`, but without it, we might get some memory leaks!


bwetherfield:
Ah, there are also a lot of inter-pitch class adjacencies to implement, so I can't guarantee this will fix everything!

Sources/SpelledPitch/PitchSpeller/Wetherfield/DirectedGraphScheme.swift
```diff
@@ -16,4 +16,8 @@ struct DirectedGraphScheme <Node>: DirectedGraphSchemeProtocol, UnweightedGraphS
     init (_ contains: @escaping (Edge) -> Bool) {
         self.contains = contains
     }
+    
+    func contains(from start: Node, to end: Node) -> Bool {
```
jsbean: I wonder if `containsEdge(from: Node, to: Node) -> Bool` might flow a little naturally.

Sources/SpelledPitch/PitchSpeller/Wetherfield/FlowNetwork.swift
```diff
@@ -51,7 +51,7 @@ extension FlowNetwork {
             }
         }
     }
-    
+
```
jsbean: Thank you Xcode?

Sources/SpelledPitch/PitchSpeller/Wetherfield/FlowNetwork.swift
```diff
@@ -60,11 +60,25 @@ extension FlowNetwork {
...
+        for edge in edges where !adjacencyScheme.contains(from: edge.a, to: edge.b) {
             remove(edge)
         }
     }
+    
+    mutating func mask <Scheme: WeightedGraphSchemeProtocol> (_ weightScheme: Scheme) where
+        Scheme.Node == Node,
+        Scheme.Weight == Weight
+    {
+        for edge in edges {
```
jsbean: Nit: I generally like the "cuddled" `else`:

```Swift
if let scalar = weightScheme.weight(from: edge.a, to: edge.b) {
    updateEdge(edge) { $0 * scalar }
else {
    remove(edge)
}
```

jsbean: (As it tightens and aligns the two conditional operations.)

Sources/SpelledPitch/PitchSpeller/Wetherfield/WeightedGraphSchemeProtocol.swift
```diff
@@ -21,25 +23,84 @@ extension WeightedGraphSchemeProtocol {
...
+extension WeightedGraphSchemeProtocol where Self: DirectedGraphSchemeProtocol, Weight: Numeric {
+    
+    static func * <Scheme> (lhs: Self, rhs: Scheme) -> Self where
+    Scheme: WeightedGraphSchemeProtocol,
+    Scheme.Node == Node,
+    Scheme.Weight == Weight
     {
-        return WeightCarrying.build(from: self).pullback(f)
+        return Self { edge in
+            guard
```
jsbean: Here, I like to have the `guard` and the `else` aligned:

```Swift
guard 
    let lweight = lhs.weight(edge),
    let rweight = rhs.weight(from: edge.a, to: edge.b)
else {
     return nil
}
```

jsbean: However, this could be shortened to:

```Swift
return lhs.weight(edge) ?? rhs.weight(from edge.a, to edge.b)
```

Sources/SpelledPitch/PitchSpeller/Wetherfield/WeightedGraphSchemeProtocol.swift
```diff
@@ -21,25 +23,84 @@ extension WeightedGraphSchemeProtocol {
...
+                let rweight = rhs.weight(from: edge.a, to: edge.b) else { return nil }
+            return lweight * rweight
+        }
+    }
+    
+    static func * <Scheme> (lhs: Self, rhs: Scheme) -> Self where
+        Scheme: UnweightedGraphSchemeProtocol,
+        Scheme.Node == Node
+    {
+        return Self { edge in
```
jsbean: I feel that a couple branches of logic could be swept away here. It would require adding this (which has been discussed for the stdlib [here](https://forums.swift.org/t/add-filter-and-ifsome-then-to-optional/8143/29)):

```Swift
extension Optional {
    func filter (_ predicate: (Wrapped) -> Bool) -> Optional {
        guard case let .some(wrapped) = self else { return nil }
        return predicate(wrapped) ? wrapped : nil
    }
}
```

We could then rewrite this as:

```Swift
return lhs.weight(edge).filter { _ in rhs.contains(from: edge.a, to: edge.b) }
```

Without adding functionality, we could pivot this like:

```Swift
return rhs.contains(from: edge.a, to: edge.b) ? lhs.weight(edge) : nil
```

Sources/SpelledPitch/PitchSpeller/Wetherfield/WeightedGraphSchemeProtocol.swift
```diff
@@ -21,25 +23,84 @@ extension WeightedGraphSchemeProtocol {
...
+    
+    static func * <Scheme> (lhs: Scheme, rhs: Self) -> Self where
+        Scheme: UnweightedGraphSchemeProtocol,
+        Scheme.Node == Node
+    {
+        return rhs * lhs
+    }
+    
+    static func + (lhs: Self, rhs: Self) -> Self {
+        return Self { edge in
```
jsbean: This might be tighter as:

```Swift
switch (lhs.weight(edge), rhs.weight(edge)) {
    case (nil,nil): return nil
    case let (lhs?, rhs?): return lhs + rhs
    case let (lhs, nil): return lhs
    case let (nil, rhs): return rhs
}
```

Or:

```Swift
return lhs.weight(edge).flatMap { lhs in rhs.weight(edge).map { rhs in lhs * rhs }
```

This seems like a perfect place for treating `Optional` as a monad, and using custom operators…

bwetherfield: Cool!

Sources/SpelledPitch/PitchSpeller/Wetherfield/DirectedGraphScheme.swift
bwetherfield: This may be nicer, I agree. What do you think about `weight(from:to:)`?

Sources/SpelledPitch/PitchSpeller/Wetherfield/FlowNetwork.swift
bwetherfield: !

Sources/SpelledPitch/PitchSpeller/Wetherfield/FlowNetwork.swift
bwetherfield: Presume you mean 
```swift
if let scalar = weightScheme.weight(from: edge.a, to: edge.b) {
    updateEdge(edge) { $0 * scalar }
} else {
    remove(edge)
}
```
which I am on board with too!

Sources/SpelledPitch/PitchSpeller/Wetherfield/WeightedGraphSchemeProtocol.swift
bwetherfield: Cool!

Sources/SpelledPitch/PitchSpeller/Wetherfield/WeightedGraphSchemeProtocol.swift
bwetherfield: whoa!

Sources/SpelledPitch/PitchSpeller/Wetherfield/WeightedGraphSchemeProtocol.swift
bwetherfield: hold on, but does this capture the logic of when they are both non-nil ?

Sources/SpelledPitch/PitchSpeller/Wetherfield/WeightedGraphSchemeProtocol.swift
bwetherfield: At present, I am going with 
```swift
            lhs.weight(edge).flatMap { weight in
                rhs.containsEdge(from: edge.a, to: edge.b) ? weight : nil
            }
```

Sources/SpelledPitch/PitchSpeller/Wetherfield/FlowNetwork.swift
jsbean: “Do as I mean, not as I say” 🤦🏻‍♂️

bwetherfield: haha!

jsbean:
This is lookin' great! Goin' in.

PR #176, opened 2018-11-08
Add more dyad tests

jsbean:
This PR adds a few more dyad tests, along with some machinery to aid them in their journey to green-ness (`bigM`, etc.).


Tests/SpelledPitchTests/PitchSpellerTests.swift
```diff
@@ -264,4 +264,48 @@ class PitchSpellerTests: XCTestCase {
...
+
+    func testSpelledOneThreeOverDNatural() {
+        let pitches: [Int: Pitch] = [0: 61, 1: 63]
+        let pitchSpeller = PitchSpeller(pitches: pitches, parsimonyPivot: Pitch.Spelling(.d))
+        let result = pitchSpeller.spell()
+        let expected: [Int: SpelledPitch] = [
+            0: SpelledPitch(Pitch.Spelling(.c, .sharp), 4),
+            1: SpelledPitch(Pitch.Spelling(.d, .sharp), 4)
+        ]
+        XCTAssertEqual(result, expected)
```
bwetherfield: I would also consider `[0: SpelledPitch(Pitch.Spelling(.d, .flat), 4), 1: SpelledPitch(Pitch.Spelling(.e, .flat), 4)]` to be a valid spelling over `.d` ... ? Can we instead check for predicates like `modifiersInSameDirection` and `noDoubles` / ("parsimony condition") ?

bwetherfield: These predicates could also be helpful for automatic test generation, since they should be true in every dyad and simpler cases with more pitches. What do you think, @jsbean? Where would they live ?

jsbean: I like the `modifiersInSameDirection`, and is in fact similar to a construct that was in place before.

However, is `modifiersInSameDirection` more an expression of "interval optimization" rather than "notational parsimony"? We have a lot of `IntervalDescriptor` machinery that can be very useful for this.

Could the "no doubles" condition be slightly more-generally represented as average line-of-fifths distance?

jsbean: > These predicates could also be helpful for automatic test generation, since they should be true in every dyad and simpler cases with more pitches. What do you think, @jsbean? Where would they live ?

That's a great question. We could create a `Fixtures` directory in the `SpelledPitchTests`. We could start there, and be scrappy with it. Then, if we think these operations would be more widely applicable, we can wash 'em off and move them upstream.

jsbean:
We are getting some very interesting failures!

`<0,1>` / `d nat` = `<b#,b##>`

jsbean:
Another issue: 

`<1,5>` / `d nat` => `<b##,e#>` (expected: `<d flat, f>`)

Sources/SpelledPitch/PitchSpeller/Wetherfield/WeightedGraphSchemeProtocol.swift
```diff
@@ -21,25 +23,70 @@ extension WeightedGraphSchemeProtocol {
...
+    }
+}
+
+extension WeightedGraphSchemeProtocol {
+    
+    /// - Returns: rhs weight function whenever the lhs is missing the edge in question.
+    /// This function is non-commutative.
+    static func + (lhs: Self, rhs: Self) -> Self {
+        return Self { edge in
+            lhs.weight(edge) ?? rhs.weight(edge)
```
jsbean: Are we supposed to be adding here if they are both present?

jsbean:
I'm going to fix these conflicts.

jsbean:
(The failing tests are as before.)

Sources/SpelledPitch/PitchSpeller/Wetherfield/WeightedGraphSchemeProtocol.swift
bwetherfield: I don't want to add in cases where they are both non-`nil`... throw an error?

jsbean: This looks great, then!

PR #177, opened 2018-11-08
[WIP / NFC / PitchSpeller] Streamline WeightedGraphSchemeProtocol * and + implementations

jsbean:
This PR ties down the implementations of `WeightedGraphSchemeProtocol` `*` and `+`.

This is to be seen from the `bigM` branch.


jsbean:
Oh, here, let me break all of the tests all at once 🥇 .

Sources/SpelledPitch/PitchSpeller/Wetherfield/WeightedGraphSchemeProtocol.swift
```diff
@@ -38,51 +38,35 @@ extension WeightedGraphSchemeProtocol {
 extension WeightedGraphSchemeProtocol where Self: UndirectedGraphSchemeProtocol, Weight: Numeric {
     
     static func * (lhs: Self, rhs: Self) -> Self {
-        return Self { edge in
-            guard let lweight = lhs.weight(edge), let rweight = rhs.weight(edge) else { return nil }
-            return lweight * rweight
-        }
+        return Self { edge in lhs.weight(edge) ?? rhs.weight(edge) }
```
bwetherfield: Does this multiply if they are both non-nil ?

Sources/SpelledPitch/PitchSpeller/Wetherfield/WeightedGraphSchemeProtocol.swift
```diff
@@ -38,51 +38,35 @@ extension WeightedGraphSchemeProtocol {
...
-        return Self { edge in
-            if let lweight = lhs.weight(edge) {
-                if let rweight = rhs.weight(edge) { return lweight + rweight }
-                return lweight
-            } else {
-                guard let rweight = rhs.weight(edge) else { return nil }
-                return rweight
-            }
+        return Self.init { edge in
+            return lhs.weight(edge).flatMap { lhs in rhs.weight(edge).map { rhs in lhs + rhs } }
```
bwetherfield: If `rhs.weight(edge)` is `nil`, don't we get `nil` instead of `lhs` ?

Sources/SpelledPitch/PitchSpeller/Wetherfield/WeightedGraphSchemeProtocol.swift
jsbean: Nope. Not one bit. V wrong.

Sources/SpelledPitch/PitchSpeller/Wetherfield/WeightedGraphSchemeProtocol.swift
jsbean: Yep. Wrong here too!

Sources/SpelledPitch/PitchSpeller/Wetherfield/WeightedGraphSchemeProtocol.swift
bwetherfield: I actually ended up using this though for a (non-commutative) addition in the other branch...

Sources/SpelledPitch/PitchSpeller/Wetherfield/WeightedGraphSchemeProtocol.swift
bwetherfield: Used this one for multiplication in the other branch, where I think this pattern works!

Sources/SpelledPitch/PitchSpeller/Wetherfield/WeightedGraphSchemeProtocol.swift
jsbean: Very resourceful: "One person's trash is another person's treasure."

jsbean:
So, this was a bust :).

PR #184, opened 2018-11-12
Harmonize with Structure @ 0.19.4

jsbean:
This PR uses the new `AdditiveGroup` API from `Structure/Algebra`, as well as the constraint-less `GraphProtocol.Weight` associated type.

It also adds tests for `WeightLabel`, as well as some various tidying up.


jsbean:
@bwetherfield, feel free to merge this into the `weight-label` branch if/when you so desire.

PR #194, opened 2018-11-13
Use test helpers to clean up PitchSpellerTests

jsbean:


jsbean:
Ok, we've got tests for all EDO12 dyads, with the seven following errors:

```
<1,10> are spelled as [b b, c #], not as the expected [c #, a #]
<3,4> are spelled as [e, e b], not as the expected [d #, e]
<3,6> are spelled as [f #, e b], not as the expected [d #, f #]
<3,11> are spelled as [b, e b], not as the expected [d #, b]
<5,6> are spelled as [f #, f], not as the expected [f, g b]
<6,10> are spelled as [b b, f #], not as the expected [f #, a #]
<10,11> are spelled as [b, b b], not as the expected [a #, b]
```

PR #199, opened 2018-11-16
Add default weight "classes"

jsbean:
This PR replaces magic numbers in initial `FlowNetwork` hook-up and in tests. Instead, boxing weight classes are used 👊.

This could get wrapped up into an `enum` for better namespacing.


bwetherfield:
Can this go straight into master (as well as the additional tests)? Minimizing red-onto-red travis scenarios

jsbean:
I could probably at least cherry pick 6ec7b82c17449607fe104ca122e7fbdb75eeec41.

bwetherfield:
I do think maybe I can skim number comparisons faster than weight-class comparisons but opinions may differ on that!

bwetherfield:
The magic numbers are definitely fairly arbitrary, so it makes sense to want to change em

jsbean:
Yeah, I just want to avoid the point when we change what those magic numbers are on the source side, then all of a sudden the flownetwork construction tests fail and it will feel muddy.

But yeah, if you have another analogy, I'm all ears :).

bwetherfield:
It's kind of too perfect seeing as we are dealing with "weights" anyway!

bwetherfield:
tough to beat that

Issue #185, opened 2018-11-13
[Discussion/ PitchSpeller] Hook up FlowNetwork as we want it rather than masking on init


jsbean:
cc @bwetherfield.

We have kind of discussed this before in #152, but I thought we could move the discussion of this particular aspect of the problem to one place. 

Currently in `PitchSpeller`, we start out by hooking up a `FlowNetwork` into a mostly-complete graph. In reality, we don't need that. As such, we spend a lot of firepower to filter out the edges we don't want, and to modify the weights according to out preferences.

Can we instead just hook up the graph how we want it (at least for initial preparation)? I am thinking we can still do this with a single quadratic pass (each node still has to look at each other node), but then we compose each of the preference-functions (`connectUpToDown`, `bigMAdjacency`, etc.) into a single function which is called at every comparison.

I think this may speed things up, and make the process feel a bit more like we are cutting the problem at its joints.

jsbean:
All of the masking machinery is of course still very valuable to injecting adjacency info, etc. into the `FlowNetwork`.

PR #202, opened 2018-11-16
[Proof of Concept / PitchSpeller] Dumb down FlowNetwork construction

jsbean:
This PR is an approach to #185. It reimplements the current FP-style `FlowNetwork` masking with a dumber, imperative style.

In this case, we don't start with a near-complete graph only to mask away edges. Instead, we build up the graph, only connecting the nodes how we want 'em.


jsbean:
The test failures are the same seven as #176.

Tests/SpelledPitchTests/PitchSpellerTests.swift
```diff
@@ -352,11 +359,33 @@ class PitchSpellerTests: XCTestCase {
...
+        switch self {
+        case .source:
+            return "source"
+        case .sink:
+            return "sink"
+        case .internal(let index):
+            return "\(index)"
+        }
+    }
+}
```
bwetherfield: These are great! Let's actually merge some of these changes in... ?

jsbean: Just want the `CustomStringConvertible` bits?

Sources/SpelledPitch/PitchSpeller/Wetherfield/Wetherfield.swift
```diff
@@ -116,37 +73,66 @@ extension PitchSpeller {
...
+                    let pcA = pitch(a)!.class
+                    let pcB = pitch(b)!.class
+                    let pair = UnorderedPair(Cross(pcA,a.tendency),Cross(pcB,b.tendency))
+                    internalEdgeLookup[pair].map { weight in
+                        flowNetwork.insertEdge(from: a, to: b, weight: weight)
+                    }
+                default:
+                    continue
+                }
+            }
```
bwetherfield: This all looks on the money to me 👍

Tests/SpelledPitchTests/PitchSpellerTests.swift
bwetherfield: Yeah, I have a feeling those will be very handy when tests fail

bwetherfield: Do they show up in tests?

jsbean: When they fail, yes. Without the `CustomStringConvertible` its like 30 pages long 😭 .

jsbean: K, I'll do another PR with them in the right place.

bwetherfield: And yes, I have read through those failures in times of bemusement!

bwetherfield: I usually end up copying into `vim` and diving into the world of selective deletion, but this is the right way to go about things (note to self)

PR #203, opened 2018-11-17
Refine Lookup naming in Pitch.Spelling.Category and TendencyConverting

jsbean:
I now see the importance of the naming here. This PR renames the `Bimap` types on `PitchSpellingCategoryProtocol` and `TendencyConverting`. `DirectionCodec` is renamed to `DirectionLookup`.

They are now called `directionToModifier` and `tendenciesToDirection`.

@bwetherfield, lemme know if that makes sense, or no.


bwetherfield:
perf!

PR #204, opened 2018-11-17
Conform FlowNode to CustomStringConvertible

jsbean:


jsbean:
Hmm, I've accumulated some git trash.

PR #206, opened 2018-11-19
FP and Imperative Coexistence

jsbean:
This PR adds the goodies from #202, but without removing the FP universe.

This way we the FP and imperative universes can live in harmony.

In a few cases, I got rid of very long generic labels, letting the compiler infer the types.


jsbean:
This does much less than I thought. All the good bits are already on `master`. Wahoo, gonna close!

PR #207, opened 2019-04-10
Inverse problem

bwetherfield:
Implements `InverseSpeller`, which takes a set of spelled pitches and outputs a set of `weights` that determine edge weights by `Pitch.Class`.


Sources/SpelledPitch/PitchSpeller/Wetherfield/Pitch.Spelling.Category.swift
```diff
@@ -10,7 +10,7 @@ import Math
 import Pitch
 
 /// Interface for the six pitch spelling categories.
-protocol PitchSpellingCategoryProtocol {
+protocol PitchSpellingCategoryProtocol: TendencyConverting {
```
jsbean: I initially wanted these to be orthogonal concepts (`PitchSpellingCategoryProtocol` and `TendencyConverting`). But that resulted in some weird awkwardness that is now cleaned up downstream as a result of this change.

So, I'm all for it. Maybe down the line we can break it up into more modular pieces if the necessity actually emerges.

Sources/SpelledPitch/PitchSpeller/Wetherfield/SpellingInverter.swift
```diff
@@ -0,0 +1,423 @@
...
+    typealias AssignedEdge = OrderedPair<PitchSpeller.AssignedNode>
+    typealias UnassignedEdge = OrderedPair<PitchSpeller.UnassignedNode>
+    typealias PitchedEdge = UnorderedPair<FlowNode<Cross<Pitch.Class,Tendency>>>
+}
+
+extension SpellingInverter {
+    
+    // MARK: - Initializers
+    
+    init(spellings: [Int: Pitch.Spelling], parsimonyPivot: Pitch.Spelling = .init(.d)) {
```
jsbean: With the added top level static properties on `Pitch.Spelling`, you can now use:

```Swift
init(spellings: [Int: Pitch.Spelling], parsimonyPivot: Pitch.Spelling = .d) {
```

Sources/SpelledPitch/PitchSpeller/Wetherfield/SpellingInverter.swift
```diff
@@ -0,0 +1,423 @@
...
+        let sameIntEdgesScheme: DirectedGraphScheme<PitchSpeller.UnassignedNode> =
+            sameIntsScheme * connectSameInts
+        
+        let specificSourceScheme: DirectedGraphScheme<PitchSpeller.UnassignedNode> =
+            sourceEdgeLookupScheme.pullback(nodeMapper)
+        
+        let specificSinkScheme: DirectedGraphScheme<PitchSpeller.UnassignedNode> =
+            sinkEdgeLookupScheme.pullback(nodeMapper)
+        
+        let maskScheme: DirectedGraphScheme<PitchSpeller.AssignedNode> =
```
jsbean: For clarity, I might just split this out into two expressions:

```Swift
        let allSchemes: DirectedGraphScheme<PitchSpeller.AssignedNode> = [
                specificEdgeScheme,
                sameIntEdgesScheme,
                specificSourceScheme,
                specificSinkScheme
            ]
        let maskScheme = allSchemes
            .reduce(DirectedGraphScheme { _ in false }, +)
            .pullback({ $0.unassigned })
```

Sources/SpelledPitch/PitchSpeller/Wetherfield/SpellingInverter.swift
```diff
@@ -0,0 +1,423 @@
...
+    }
+}
+
+extension SpellingInverter {
+    
+    // MARK: - Computed Properties
+    
+    /// - Returns: A concrete distribution of weights to satisfy the weight relationships delimited by
+    /// `weightDependencies` or `nil` if no such distribution is possible, i.e. there are cyclical
+    /// dependencies between edge types. In the latter case, the spellings fed in are *inconsistent*.
```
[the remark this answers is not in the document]
bwetherfield: Awesome idea. I'm not a wizard with things like `precondition` but this looks like a good use case. It's a DFS so O(V+E). _May_ ultimately be a public property, like if you try to generate some edge weights from a bunch of spelling preferences and it wants to tell you that "this is an inconsistent set of spellings, try again" kind of thing

Sources/SpelledPitch/PitchSpeller/Wetherfield/SpellingInverter.swift
bwetherfield: perfect

Sources/SpelledPitch/PitchSpeller/Wetherfield/SpellingInverter.swift
bwetherfield: whoa

Sources/SpelledPitch/PitchSpeller/Wetherfield/Pitch.Spelling.Category.swift
bwetherfield: Yeah this was probably one of those reckless hacking moments so that things would work with the `Bimap`s I needed to implement. Sorry about messing with the API vision!

Sources/SpelledPitch/PitchSpeller/Wetherfield/SpellingInverter.swift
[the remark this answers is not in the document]
bwetherfield: Oh wait but it's O(V+E) where the V is an _edge set_! So it's like O(Edges + Edges of Edges). But that's worst case - I think the graph should always be pretty sparse. I have to check that.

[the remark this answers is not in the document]
jsbean: Whoops, and I think it should actually be:

```Swift
precondition(!findCycle(dependencies))
```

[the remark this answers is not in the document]
bwetherfield: Right!

Sources/SpelledPitch/PitchSpeller/Wetherfield/SpellingInverter.swift
```diff
@@ -0,0 +1,430 @@
...
+                    return result + recursiveReducer(&weights, (key: edge, value: dependencies[edge]!))
+                }
+                weights[dependency.key] = weight
+                return weight
+            }
+            
+            let _ = recursiveReducer(&weights, dependency)
+        }
+        
+        return dependencies.reduce(into: [PitchedEdge: Double](), dependeciesReducer)
```
jsbean: Not sure if this is that helpful, but you can do:

```Swift
return dependencies.reduce(into: [:], dependeciesReducer)
```

The types are inferred just fine.

Sources/SpelledPitch/PitchSpeller/Wetherfield/SpellingInverter.swift
```diff
@@ -0,0 +1,430 @@
...
+    
+    // MARK: - Computed Properties
+    
+    /// - Returns: A concrete distribution of weights to satisfy the weight relationships delimited by
+    /// `weightDependencies` or `nil` if no such distribution is possible, i.e. there are cyclical
+    /// dependencies between edge types. In the latter case, the spellings fed in are *inconsistent*.
+    /// Weights are parametrized by `Pitch.Class` and `Tendency` values.
+    var weights: [PitchedEdge: Double] {
+        
+        let dependencies = pitchedDependencies
```
jsbean: Is this necessary? I tried substituting `dependencies` with `pitchedDependencies` and all the tests passed.

In this case, `pitchedDependencies` is a Copy-On-Write `Dictionary`, so the copy is made for you, only when needed.

jsbean:
@bwetherfield, I'm cool with this if you are cool with this!

Sources/SpelledPitch/PitchSpeller/Wetherfield/SpellingInverter.swift
bwetherfield: Does a computed property get recomputed every time it is called? I think that's why I stashed it

jsbean: Oooooh boy, good point. I must have overlooked that it has a non-trivial implementation. Not sure what I was thinking.

An idiom which I haven't done a good job of following myself (though perhaps I aspire to), is to use functions for any properties which are not O(*1*), so maybe we should use:

```Swift
func pitchedDependencies() -> [Pitch: Double]
```

jsbean: Sorry about that 🤦‍♂

bwetherfield: I like that convention!

PR #208, opened 2019-04-11
Generate linux tests

jsbean:
This PR simply generates linux tests to ensure that there is no funny business when testing on Linux.

This can be done with `swift package --generate-linuxmain`.


bwetherfield:
Whoa!

PR #212, opened 2019-05-23
Derive edge weights from graphs with cyclic dependencies

bwetherfield:
Until this pull request, the inverse problem solver only works when the input graph has no cyclic dependencies (hence the `findCycle` precondition). In reality, the problem of cycles in the dependencies of edge weights can be solved by fixing all the edge weights that are cyclically dependent as _equal_. This is achieved with the apparatus of this pull request by clumping all strongly connected components of the input dependency structure (just a directed graph on the edges) using Tarjan's algorithm (so that the clumped graph is a DAG), applying the inverse solver strategy on the _clumped_ graph and then "unwrapping" the clumped graph - by assigning the weight of each clump to the edges that that clump contains.


bwetherfield:
This pull request contains a fair bit of algorithmic machinery for "adjacency lists" of the form `[Node: Set<Node>]`. I'd like to formalize a data structure of the form 
```Swift
struct AdjacencyList<Node: Hashable> {
    var adjacencies: [Node: Set<Node>]
}```
and move all this machinery down to a def inside of `Structure`.

**Comment by jsbean** [comment]
This is looking and feeling good to me. Do you want to merge this in, then update `Structure`, then update here, or do we want to do the `Structure` update first, then update within this PR, then merge?

File: Sources/SpelledPitch/PitchSpeller/Wetherfield/SpellingInverter.swift [hunk 1]
```diff
@@ -32,6 +32,13 @@ extension SpellingInverter {
     
     // MARK: - Initializers
     
+    init(spellings: [Pitch.Spelling], parsimonyPivot: Pitch.Spelling = .d) {
+        let dict = spellings.enumerated().reduce(into: [Int: Pitch.Spelling]()) {
```
jsbean: Nit: you can get away with `.reduce(into: [:]) {` here, but no biggie.

Sources/SpelledPitch/PitchSpeller/Wetherfield/SpellingInverter.swift
```diff
@@ -100,6 +109,44 @@ extension SpellingInverter {
         return pitchedDependencies.reduce(into: [:], dependeciesReducer)
     }
     
+    func generateWeightsFromCycles (_ dependencies: [PitchedEdge: Set<PitchedEdge>])
+        -> [PitchedEdge: Double] {
+        let clumper = findStronglyConnectedComponent(dependencies)
+        let directedAcyclicGraph = groupNodes(dependencies, clumper)
```
jsbean: Not suggesting we do this, but would it ever be feasible to implement a DAG at the type level?

Sources/SpelledPitch/PitchSpeller/Wetherfield/SpellingInverter.swift
```diff
@@ -100,6 +109,44 @@ extension SpellingInverter {
...
+            _ dependency: (key: Node, value: Set<Node>)
+            )
+        {
+            func recursiveReducer (
+                _ weights: inout [Node: Double],
+                _ dependency: (key: Node, value: Set<Node>)
+                ) -> Double
+            {
+                let weight = dependency.value.reduce(1.0) { result, edge in
+                    if weights[edge] != nil { return result + weights[edge]! }
```
jsbean: You could do:

```Swift
if let edgeWeight = weights[edge] { return edgeWeight } 
```

Sources/SpelledPitch/PitchSpeller/Wetherfield/SpellingInverter.swift
```diff
@@ -160,24 +207,125 @@ extension SpellingInverter {
     
     // MARK: - Instance Methods
     
+    mutating func partition (_ indices: [Int: Int]) {
+        let adjacencyScheme = GraphScheme<FlowNode<Int>> { edge in
+            switch (edge.a, edge.b) {
+            case let (.internal(a), .internal(b)):
+                return indices[a]! == indices[b]!
```
jsbean: You can get away with

```Swift
return indices[a] == indices[b]
```

as the `Optional` type gets `Equatable` conformance for free.

bwetherfield: nice

Sources/SpelledPitch/PitchSpeller/Wetherfield/SpellingInverter.swift
bwetherfield: Interesting. The only way I could think of doing this would be storing paths (as a kind of "ordered set"), which would probably be quite space inefficient, but it would be nice if this were possible in a way that was not too costly

Sources/SpelledPitch/PitchSpeller/Wetherfield/SpellingInverter.swift
bwetherfield: nice

Sources/SpelledPitch/PitchSpeller/Wetherfield/SpellingInverter.swift
bwetherfield: I think I will do this and then do away with all the $0's and $1's for clarity

bwetherfield:
> This is looking and feeling good to me. Do you want to merge this in, then update `Structure`, then update here, or do we want to do the `Structure` update first, then update within this PR, then merge?

I think the implementation of `findCycle` up here is a little faulty - might not work on a trivial self-loop case, so I might add a failing test to illustrate... the best way to fix the fault may be to pull in the updated `Structure`, since I have it working down there. What do you think?

Sources/SpelledPitch/PitchSpeller/Wetherfield/SpellingInverter.swift
bwetherfield: Ah, so because of the later usage, the compiler complained about not knowing the type of dict, so do you think it is better to have `dict: [Int: Pitch.Spelling]` or the explicit `reduct(into: [Int: Pitch.Spelling]())`?

bwetherfield: For now I am going with the former

